PR #10, opened 2026-07-10
fix(otp): parse KeeOTP TOTP query fields

### LINKED ISSUE #9: fix: recognize KeeOTP TOTP query fields
## Problem

KeeForge does not recognize some valid KeePass OTP entries. In particular, the KeeOTP plugin stores TOTP configuration in the custom `otp` field as a query string instead of an `otpauth://` URI:

```text
otp=key=<secret>&type=TOTP&step=30&size=6&encoding=Base32&otpHashMode=SHA1
```

When this value is opened in KeeForge, no `totpConfig` is created and the OTP section is not shown.

## First implementation slice

This change targets KeeOTP **TOTP** compatibility:

- custom fields named exactly `otp`, `OTP`, or `Otp`;
- values beginning with `key=`;
- URL-query decoding;
- strict support for the KeeOTP encodings `Base32`, `Base64`, `Hex`, and `UTF8`;
- explicit validation of `type=TOTP`, `step`, `size`, and `otpHashMode`;
- no decoder fallback or heuristic interpretation;
- preservation of existing OTP source precedence and behavior;
- regression coverage for malformed values, duplicate parameters, encoding boundaries, and lowercase Base32 keys.

HOTP is intentionally deferred to a separate change because it requires a mutable counter, persistence, conflict handling, and one-shot generation semantics.

## Compatibility evidence

The KeeOtp2 implementation confirms:

- legacy data is stored in the `otp` field;
- supported parameters include `key`, `type`, `step`, `size`, `encoding`, `counter`, and `otpHashMode`;
- KeeOtp2 accepts Base32 keys in lowercase;
- HOTP uses a separate `counter` lifecycle and is outside this first slice.

Reference: [KeeOtp2](https://github.com/tiuub/KeeOtp2/blob/17a8545cf5906df693e748b99a347e96425df10d/KeeOtp2/OtpAuthUtils.cs)

## Verification

The implementation includes focused XCTest coverage. Local verification is pending because `xcodebuild` is unavailable on Linux. The repository CI workflow runs on `macos-15` with Xcode and can validate the change there.

## Follow-up candidates

- HOTP support with persistent `counter` / `HmacOtp-Counter`.
- OTP URIs stored in the standard `URL` field.
- Additional URI schemes such as `otp://`, after establishing a verified compatibility baseline.
- Raw Base32-only values without explicit KeeOTP query metadata.

ftorga:
Closes #9

## Summary
- Parse KeeOTP TOTP query payloads from supported entry fields.
- Decode declared secret encodings safely and feed existing TOTP consumers.
- Keep HOTP out of this slice.

## Changes
| File | Change |
|------|--------|
| `KeeForge/Models/EncryptedValue.swift` | Tightened secret handling for OTP payload decoding |
| `KeeForge/Models/Entry.swift` | Added KeeOTP-aware TOTP wiring |
| `KeeForge/Models/KDBXParser.swift` | Parsed KeeOTP TOTP query fields |
| `KeeForge/Models/TOTPGenerator.swift` | Adapted TOTP generation inputs |
| `KeeForgeTests/KDBXParserTests.swift` | Added parser coverage |
| `KeeForgeTests/TOTPGeneratorTests.swift` | Added generator coverage |

## Test Plan
- [x] Unit tests added for parser and generator behavior
- [x] Existing TOTP flows preserved
- [x] HOTP intentionally deferred

## Checklist
- [x] Linked issue #9
- [x] Conventional commit format
- [x] No `Co-Authored-By` trailers
- [ ] Repository-specific labels are not configured in this fork/upstream pair yet


KeeForge/Models/KDBXParser.swift
```diff
@@ -1666,9 +1667,100 @@ private class EntryBuilder {
...
+        )
+    }
+
+    private func decodeKeeOTPSecret(_ key: String, encoding: String) -> Data? {
+        switch encoding.lowercased() {
+        case "base32":
+            guard let normalized = normalizeStrictUnpaddedBase32(key) else { return nil }
+            return TOTPGenerator.base32Decode(normalized).flatMap { $0.isEmpty ? nil : $0 }
+        case "base64":
+            return Data(base64Encoded: key).flatMap { $0.isEmpty ? nil : $0 }
```
crazytan: **[P1] Make Base64 decoding strict.** `Data(base64Encoded:)` accepts the over-padded `JBSWY3DP====`, so this branch creates a `TOTPConfig` even though the declared Base64 spelling is malformed. This is why the PR's own `testKeeOTPSupportsEachDeclaredEncodingWithoutFallback` fails. Please validate canonical Base64 characters/padding (or round-trip decoded bytes to the canonical representation) before accepting the secret.

KeeForge/Models/Entry.swift
```diff
@@ -120,12 +120,21 @@ struct KPEntry: Identifiable, Sendable {
 /// TOTP configuration extracted from KeePass entry
 struct TOTPConfig: Sendable {
     let secret: EncryptedValue
+    /// Pre-decoded secret bytes for formats whose declared encoding is not Base32.
+    let decodedSecret: EncryptedValue?
```
crazytan, line 143: **[P1] Preserve the decoded secret/source encoding through edits.** `EntryDraftPayload` and `DatabaseDraft.makeTOTPConfig` do not carry `decodedSecret`, so an unrelated entry edit reconstructs Base64/Hex/UTF-8 secrets as ordinary Base32 and breaks TOTP generation. For uppercase `OTP`/`Otp`, serialization also writes conflicting `TimeOtp-*` fields alongside the original custom field, and those fields take precedence on reload. Please preserve the KeeOTP source/encoding or decoded bytes through the edit and serialization pipeline, with an edit/save regression test.

crazytan (changes requested):
Thanks for the focused compatibility work. I found two blocking correctness issues that should be addressed before merge. I also ran the relevant unit-test slice locally; it completed with one failure in `KDBXParserTests.testKeeOTPSupportsEachDeclaredEncodingWithoutFallback()` (the over-padded Base64 case). `TOTPGeneratorTests` and `EncryptedValueTests` passed. There are currently no GitHub workflow runs/statuses attached to this commit.

Please also add the required `CHANGELOG.md` entry under `## Unreleased` for this bug fix.

KeeForge/ViewModels/EntryEditViewModel.swift
```diff
@@ -224,6 +234,11 @@ final class EntryEditViewModel {
 
         return EntryDraftPayload.TOTPConfiguration(
             secret: secret,
+            decodedSecret: secret == originalSnapshot.totpSecret.trimmingCharacters(in: .whitespacesAndNewlines)
```
crazytan: **[P1] Preserve the exact lowercase KeeOTP source when normalization changes the display secret.** For a valid lowercase `otp` query whose UTF-8 key begins or ends with percent-encoded whitespace, this comparison succeeds after trimming and retains `decodedSecret`, but the payload's `secret` is the trimmed value. `DatabaseDraft.preservedOtpURL` then compares that trimmed value with the exact original secret, returns `nil`, and serialization falls back to writing the trimmed UTF-8 text as `TimeOtp-Secret-Base32`. An unrelated edit therefore changes or invalidates the TOTP secret. I reproduced this with `key=%20leading%20and%20trailing%20&...&encoding=UTF8`. Please preserve source provenance/query independently of normalized editor text and add the lowercase-whitespace round-trip case.

crazytan (commented):
Re-review completed on `6a9da2b`. The earlier strict-Base64 issue is fixed, and the updated focused suite passes: `KDBXParserTests`, `TOTPGeneratorTests`, `EncryptedValueTests`, `KDBXRoundTripTests`, `KDBXWriterTests`, `DatabaseDraftTests`, and `EntryEditViewModelTests`.

There is one remaining blocking preservation edge case, noted inline. A temporary regression test for a lowercase `otp` UTF-8 secret with leading/trailing whitespace failed because an unrelated edit dropped `updatedEntry.otpURL`.

Merge-readiness follow-ups: the branch currently conflicts with `main` in `CHANGELOG.md`, and both PR commits need `Signed-off-by:` trailers under the repository's DCO policy. GitHub currently shows no CI workflow/status for this head.

KeeForge/ViewModels/EntryEditViewModel.swift
ftorga: Fixed in 0ae88b8. Unrelated edits now preserve the original untrimmed KeeOTP secret and exact lowercase source. The new UTF-8 leading/trailing whitespace regression failed before the fix and passes afterward, along with KDBXRoundTripTests and EntryEditViewModelTests.

KeeForge/Models/KDBXParser.swift
ftorga: Fixed in 6a9da2b. Base64 decoding now requires a canonical round trip, and malformed padding is covered by the focused parser tests.

KeeForge/Models/Entry.swift
ftorga: Fixed in 6a9da2b. Decoded KeeOTP secrets now survive the edit and draft pipeline, original uppercase sources are preserved, and edit/save round-trip coverage was added.

ftorga:
All requested changes are now addressed and the review threads are resolved. The final whitespace preservation fix is in `0ae88b8`; the focused regression, `KDBXRoundTripTests`, and `EntryEditViewModelTests` pass. DCO is green and the branch is conflict-free. Could you please re-review when convenient?

KeeForge/Models/KDBXXMLSerializer.swift
```diff
@@ -274,7 +274,8 @@ struct KDBXXMLSerializer {
             )
             knownChildCount += 1
             attachmentAnchor += 1
-        } else if let totpConfig = entry.totpConfig {
+        } else if let totpConfig = entry.totpConfig,
+                  totpConfig.decodedSecret == nil || !hasKeeOTPCustomField(entry) {
```
crazytan: **[P1] Deliberate TOTP edits of a KeeOTP entry silently corrupt the config.** The preservation logic only covers unrelated edits: when the user intentionally changes period/digits/algorithm/secret, `normalizedTOTPConfiguration()` drops `decodedSecret`, so this guard passes and the branch writes `TimeOtp-Secret-Base32` containing the raw Base64/Hex/UTF-8 string, which is not valid Base32.

Repro (encoding=Base64, key `AAEC/w==`, user changes period 30→60):
- **Lowercase `otp` source**: `preservedOtpURL` correctly returns nil, this branch emits `TimeOtp-Secret-Base32 = "AAEC/w=="`, and on reload the `TimeOtp-` path takes precedence, `base32Decode` fails, and the TOTP renders `------` — a working secret destroyed by a settings tweak.
- **Uppercase `OTP` source**: same emission, plus the stale `OTP` custom field rides through `draft.customFields` and survives. Even when the edit works locally (user pastes a new valid Base32 secret), other clients reading the KeeOTP field see the old, conflicting config.

Please rewrite or remove the KeeOTP source (`otpURL` / `OTP` custom field) when the TOTP configuration intentionally changes, rather than emitting `TimeOtp-*` alongside or instead of it, and add regression tests for a period change and a secret change on a Base64-encoded entry for both source spellings.

KeeForge/Models/KDBXParser.swift
```diff
@@ -1325,6 +1325,7 @@ final class KDBXXMLParser: NSObject, XMLParserDelegate {
                 case "URL": entry.url = currentValue
                 case "Notes": entry.notes = currentValue
                 case "otp": entry.otpURL = currentValue
+                case "OTP", "Otp": entry.customFields[currentKey] = currentValue
```
crazytan: **[P2] The secret-bearing KeeOTP query surfaces as an ordinary editable custom field.** The builder strips `TimeOtp-*`/`TOTP Seed`/`TOTP Settings` from `customFields` after extraction, but `OTP`/`Otp` are deliberately kept for round-trip preservation, and `KPEntry.displayCustomFields` filters only passkey keys. The result: the raw query — plaintext secret included — appears in entry details and in the edit screen as a normal custom field, duplicating the dedicated TOTP editor, and hand-edits to that field desync from `totpConfig`. Please filter these keys from display/edit the way `TimeOtp-*` is handled, once preservation no longer depends on them riding through the editable custom-fields list.

crazytan: **[P3] This case is redundant.** The `default` branch's `else if !currentKey.isEmpty` already stores any non-empty key into `entry.customFields`, so `OTP`/`Otp` were preserved before this change too. Drop the case (or, if you keep it for documentation value, consider collapsing the `default` branch's two identical arms, which this change makes more conspicuous).

KeeForge/Models/KDBXParser.swift
```diff
@@ -1684,9 +1685,102 @@ private class EntryBuilder {
...
+        for value in candidates where value?.hasPrefix("key=") == true {
+            if let value, let config = parseKeeOTPTOTP(value, sessionKey: sessionKey) {
+                return config
+            }
+        }
         return nil
     }
 
+    private func parseKeeOTPTOTP(_ query: String, sessionKey: SymmetricKey) -> TOTPConfig? {
+        guard let components = URLComponents(string: "https://keeotp.invalid/?\(query)") else { return nil }
```
crazytan, line 1700: **[P3] Compatibility note: `URLComponents` does not decode `+` as space.** A KeeOTP writer that space-encodes a UTF-8 key with `+` instead of `%20` would have its secret decoded literally. KeeOtp/KeeOtp2 appear to emit `%20`, so this is likely fine — just flagging as a known limitation; a one-line comment here would prevent a future "bug" report from re-litigating it.

KeeForge/Models/TOTPGenerator.swift
```diff
@@ -14,8 +14,15 @@ enum TOTPGenerator {
     }
 
     static func resolveSecret(config: TOTPConfig, sessionKey: SymmetricKey) -> ResolvedSecret? {
-        guard let secretString = try? config.secret.decrypt(using: sessionKey),
-              let secretData = base32Decode(secretString) else { return nil }
+        let secretData: Data?
+        if let decodedSecret = config.decodedSecret {
+        secretData = try? decodedSecret.decryptData(using: sessionKey)
```
crazytan: **[P3] Indentation:** the branch bodies in `resolveSecret` (lines 19, 21, 23) are flush with their `if`/`else` keywords instead of indented one level.

KeeForgeTests/KDBXParserTests.swift
```diff
@@ -197,7 +197,102 @@ final class KDBXParserTests: XCTestCase {
         }
     }
 
-    // MARK: - Group Membership Tests
+        // MARK: - KeeOTP TOTP Compatibility
```
crazytan, line 200: **[P3] Indentation:** the new KeeOTP test block (and the relocated `// MARK: - Group Membership Tests`) uses 8-space indentation, inconsistent with the rest of the file; the `parseSingleEntry`/`xmlEscape` helper bodies at the bottom of the file are also under-indented. Please normalize to the file's 4-space style.

crazytan (changes requested):
Re-review completed on `0ae88b8`. All three previously blocking issues are confirmed fixed: strict canonical Base64 round-trip validation, `decodedSecret` preservation through the edit/draft/serializer pipeline, and exact lowercase `otp` source preservation for whitespace-bearing UTF-8 secrets. I ran the focused suites locally (`KDBXParserTests`, `TOTPGeneratorTests`, `KDBXRoundTripTests`, `KDBXWriterTests`, `EntryEditViewModelTests`, `DatabaseDraftTests`, `EncryptedValueTests`) — all pass with 0 failures. DCO is green on all three code commits, the `CHANGELOG.md` conflict is resolved, and the branch is mergeable.

This round found **one new blocking issue** (noted inline on `KDBXXMLSerializer.swift`): the preservation fix protects *unrelated* edits, but a **deliberate** TOTP edit of a KeeOTP entry with a non-Base32 encoding still silently corrupts the configuration.

One repo-convention item that has no diff line to anchor to: `KeeForgeTests/README.md` designates `KeeForgeTests/KDBXCompatibilityTests.swift` as the authoritative end-to-end compatibility matrix, and AGENTS.md requires updating it (and the compatibility artifact gate, if applicable) when parser/writer compatibility changes. This PR adds a new supported OTP storage format that round-trips through edit/save, but the compatibility tests were not touched. Please add KeeOTP fixture shapes (`otp`/`OTP` sources across the four encodings) to the matrix.

Remaining inline comments are P2/P3: the secret-bearing KeeOTP query surfacing as an editable custom field, a redundant parser case, a `+`-decoding compatibility note, and indentation cleanups.

PR #17, opened 2026-07-15
Reload databases changed outside KeeForge

### LINKED ISSUE #13: Reload the database when the file has been updated
Is it possible to reload the database when a new version of the database file has been copied to the iPhone? This way, any changes would appear immediately. At the moment, I have to remove the database entry and add it again whenever something changes in the .kdbx file and I manually sync the file.

crazytan:
## Summary

- detect when an unlocked local or cloud database changes outside KeeForge
- parse and apply the refreshed database to the active session
- defer reloads during editing, saving, or unsaved changes
- bound deferred retries so persistent refresh failures cannot create a tight loop

## Root cause

Foreground refresh updated the shared AutoFill cache and identity store, but did not replace the unlocked in-memory database tree. The UI therefore remained stale after another process replaced the backing file.

## Impact

KeeForge now shows externally updated database contents after foreground refresh while protecting active edits and unsaved changes.

## Validation

- `xcodebuild test -project KeeForge.xcodeproj -scheme KeeForge -destination 'platform=iOS Simulator,name=iPhone 17 Pro' -derivedDataPath /tmp/KeeForgeIssue13Derived -only-testing:KeeForgeTests/DatabaseViewModelTests -quiet`

Closes #13


crazytan:
Closing for now. It'd be a nice feature but needs to be carefully thought out

Issue #1, opened 2026-03-26
"KDF parameter out of range" error when trying to open database


TentakillLIVE:
<img width="603" height="1311" alt="Image" src="https://github.com/user-attachments/assets/05d986f3-2996-4df5-901f-6ad299c70516" />

Database was built with latest version of KeePassXC (2.7.12)
On latest version of iOS (26.4)

crazytan:
@TentakillLIVE thanks for reporting this! Relaxed the validation in https://github.com/crazytan/KeeForge/commit/875c5163a9a9d717535ed29918adcaa9b6e0b1cf and will let you know when it's released. Happy to add you to TestFlight group as well.

TentakillLIVE:
I am interested in the TestFlight

crazytan:
Can you email your Apple ID to support@keeforge.com? once I have a new build I'll add you to it.

crazytan:
@TentakillLIVE

crazytan:
@TentakillLIVE the fix is released as v1.5.1 now, closing. Let me know if it still doesn't work

PR #27, opened 2026-07-21
test(crypto): add RFC 8439 known-answer vectors for inner-stream ChaCha20

### LINKED ISSUE #25: ChaCha20 inner stream has no known-answer tests — add RFC 8439 vectors?
Hey! I was reading through the crypto and noticed the inner-stream ChaCha20 in `KDBXCrypto.swift` (the `chacha20XOR` bit around line 376) is hand-rolled, which totally makes sense since CryptoKit only gives you ChaChaPoly and not the raw stream. No complaints there.

The thing I got stuck on: as far as I can tell it's only covered by round-trip and synthetic-fixture tests. Those don't actually catch a subtly-wrong keystream, because encrypt→decrypt cancels out even if the state setup or quarter-rounds are off — as long as they're wrong the same way both directions. So there's nothing in the suite that pins the output to the real ChaCha20 spec.

To be clear I'm not saying it's broken — I couldn't find a bug. It just isn't *proven* correct anywhere, and for a password manager that felt worth raising.

Easy fix would be to drop in the RFC 8439 reference vectors as a known-answer test. Appendix A.1 Test Vector #1 is convenient because it uses block counter 0, same as this implementation, so it goes straight through the existing entry point:

```swift
func test_chacha20_matchesRFC8439_appendixA1_vector1() throws {
    // RFC 8439, Appendix A.1, Test Vector #1
    let key = Data(repeating: 0, count: 32)
    let nonce = Data(repeating: 0, count: 12)
    // XOR-based, so encrypting 64 zero bytes gives the raw keystream.
    let zeros = Data(repeating: 0, count: 64)

    let keystream = try KDBXCrypto.encryptChaCha20(data: zeros, key: key, nonce: nonce)

    let expected = Data([
        0x76, 0xb8, 0xe0, 0xad, 0xa0, 0xf1, 0x3d, 0x90,
        0x40, 0x5d, 0x6a, 0xe5, 0x53, 0x86, 0xbd, 0x28,
        0xbd, 0xd2, 0x19, 0xb8, 0xa0, 0x8d, 0xed, 0x1a,
        0xa8, 0x36, 0xef, 0xcc, 0x8b, 0x77, 0x0d, 0xc7,
        0xda, 0x41, 0x59, 0x7c, 0x51, 0x57, 0x48, 0x8d,
        0x77, 0x24, 0xe0, 0x3f, 0xb8, 0xd8, 0x4a, 0x37,
        0x6a, 0x43, 0xb8, 0xf4, 0x15, 0x18, 0xa1, 0x1c,
        0xc3, 0x87, 0xb6, 0x69, 0xb2, 0xee, 0x65, 0x86,
    ])

    XCTAssertEqual(keystream, expected)
}
```

The other Appendix A.1 vectors would be nice too for non-zero keys and multi-block coverage — a couple of them start at a non-zero counter though, so you'd need to expose an initial-counter param on the internal entry point just for testing if you want full coverage (including the §2.4.2 sunscreen vector).

Happy to open a PR with the test if that's useful. Thanks for the app btw.

miquno:
Closes #25.

Adds RFC 8439 Appendix A.1 Test Vector #1 as a known-answer test for the hand-rolled inner-stream ChaCha20, plus a decrypt-symmetry check and a multi-block first-block check. Also registers the new file in both test targets (the project uses explicit membership, not synchronized groups, so it wouldn't compile in otherwise).

Verified locally: `Executed 3 tests, with 0 failures` on the iOS simulator. The file name and test naming are just my picks let me know if you'd prefer a different convention.

### DIFF
--- a/KeeForge.xcodeproj/project.pbxproj
+++ b/KeeForge.xcodeproj/project.pbxproj
@@ -246,6 +246,7 @@
 		7725E1109FAED809C9923CFD /* test-v1.key in Resources */ = {isa = PBXBuildFile; fileRef = 4C350259754DFA5D42D782A7 /* test-v1.key */; };
 		775F9BA8EFAEF1F361AFF6BB /* Entry.swift in Sources */ = {isa = PBXBuildFile; fileRef = AEF7A2480DF3853AFF0AF02B /* Entry.swift */; };
 		776466A679F17BCB1E7E3762 /* TwofishTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = C0D923EB4B018FB500295E74 /* TwofishTests.swift */; };
+		665A47FBEF96CB7164341DA2 /* ChaCha20Tests.swift in Sources */ = {isa = PBXBuildFile; fileRef = B87477A6BC8CBFED1E9B1A05 /* ChaCha20Tests.swift */; };
 		7773C42734CD4B651C8EF7B7 /* CredentialMatcher.swift in Sources */ = {isa = PBXBuildFile; fileRef = AB8B2C6F3F5C2509AC462D87 /* CredentialMatcher.swift */; };
 		77E08F862D8F74AE6898370A /* PasswordDisplay.swift in Sources */ = {isa = PBXBuildFile; fileRef = 10E0212432C502CF3BCFAB93 /* PasswordDisplay.swift */; };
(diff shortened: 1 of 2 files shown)


crazytan (commented):
Thanks for creating the PR!

crazytan (approved):

PR #28, opened 2026-07-21
fix(security): keep ATS enforced for first-party hosts

### LINKED ISSUE #26: Idea: keep ATS on for the app's own hosts?
Hey! Been poking around the codebase (really nice work btw) and had a small idea, totally take it or leave it.

I saw both `Info.plist`s use `NSAllowsArbitraryLoads = true`, and the comments explain why — user-entered WebDAV servers, some with crusty old ciphers, and ATS gives you no way to make an exception for hosts you don't know ahead of time. Makes total sense, that part kind of has to stay.

The only thing I noticed: since it's a blanket setting, it also relaxes ATS for the handful of hosts the app *does* know about and controls — `icons.duckduckgo.com`, `feedback.keeforge.com`, the OAuth stuff. Those are all fine modern HTTPS, so they could just keep ATS on for free.

If it seems worth it, you can list those under `NSExceptionDomains` — anything you list there gets ATS re-applied, and anything you don't still falls under arbitrary loads, so the WebDAV path keeps working exactly as-is:

```xml
<key>NSAppTransportSecurity</key>
<dict>
    <key>NSAllowsArbitraryLoads</key>
    <true/>
    <key>NSExceptionDomains</key>
    <dict>
        <key>duckduckgo.com</key>
        <dict>
            <key>NSIncludesSubdomains</key><true/>
        </dict>
        <key>keeforge.com</key>
        <dict>
            <key>NSIncludesSubdomains</key><true/>
        </dict>
        <!-- could add dropbox / microsoft login+graph here too -->
    </dict>
</dict>
```

(A bare exception dict already re-imposes the good defaults — forward secrecy, TLS 1.2, no cleartext — so nothing more needed.)

Anyway, minor thing, and the current setup isn't wrong just thought it was a cheap way to keep the protection where it doesn't cost anything. 

Happy to send a PR too if you want :D

miquno:
Closes #26.

Keeps `NSAllowsArbitraryLoads` on for user-entered WebDAV servers, but adds `NSExceptionDomains` for the app's own fixed hosts — `duckduckgo.com` (favicons), `keeforge.com` (feedback), and the Dropbox/Microsoft cloud endpoints — so ATS stays enforced for them. Anything not listed still falls under arbitrary loads, so the user-entered WebDAV path is unaffected. Includes a CHANGELOG entry.

Both `Info.plist`s pass `plutil -lint`, the app builds, and the built app's processed `Info.plist` contains the exceptions as expected.

<sub>Note: this branch briefly carried a second commit raising the timeouts in `testSaveUnderNSFileCoordinatorDoesNotDeadlockWhenOtherCoordinatorActive`, which was failing the CI gate on unrelated PRs. You've since fixed that on `main` (same 5s/30s bounds), so I dropped my commit — this PR is ATS-only again and merges cleanly.</sub>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ TODO before the first macOS release:
 
 ## Unreleased
 
+- App Transport Security now stays enforced for the app's own fixed hosts — favicons (DuckDuckGo), feedback, and the Dropbox/Microsoft cloud endpoints — via per-domain exceptions. The global arbitrary-loads allowance remains only for user-entered WebDAV servers, many of which still require legacy TLS ciphers that ATS otherwise rejects.
 - Choose which databases appear in AutoFill, get suggestions from all of them at once, and clear AutoFill suggestions in Settings.
 - Switch between databases inside the AutoFill panel.
 - Saving a database now regenerates the file header's random material — master seed, encryption IV, inner stream key, and KDF salt — on every save, matching KeePass 2.x and KeePassXC. KDF cost settings (Argon2 iterations/memory/parallelism, AES-KDF rounds) are preserved, and saved files remain fully compatible with other KeePass clients. Thanks to Kinglike1337 for the contribution.
--- a/KeeForge/Info.plist
+++ b/KeeForge/Info.plist
@@ -100,6 +100,46 @@
 		-->
 		<key>NSAllowsArbitraryLoads</key>
 		<true/>
+		<!--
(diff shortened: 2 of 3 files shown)


crazytan (commented):
Thanks for creating the PR! Can you add an entry in CHANGELOG.md as well?  It'd be great if you can also add dropbox and microsoft domains but not a requirement.

Issue #3, opened 2026-04-18
Auto-lock Timeout setting is ignored


dl97ag:
Regardless of Auto-lock Timeout setting (I’ve tried 5 mins and Never), switching apps away from KF, and back to KF immediately, results in KF locking the DB rather than displaying the opened DB (or entry being viewed).

crazytan:
Thanks for reporting. Currently, this setting is a bit confusing because the app always locks the database when switching to other apps. This auto-lock timeout setting is actually referring to how long the app stays open on the screen before it automatically locks an unlocked database.

I can add an option to keep the database open when switching to other apps. The timeout will count down from the last activity with the app. Does that work? @dl97ag

dl97ag:
Sure thing - appreciate your quick feedback!

I didn’t realize the DB could be locked during an active session. Yes, your suggestion works for me 👍🏻

crazytan:
Of course! Happy to make this quick fix. I'll let you know when it's released. And also small clarification: the timeout only starts from inactivity, not during active use (I'll double check though).

crazytan:
@dl97ag the new option should be live on 1.8.2 update now, please check it out

PR #42, opened 2026-07-27
fix(autofill): prevent biometric unlock crashes

### LINKED ISSUE #41: fix(autofill): prevent crash after biometric unlock when Quick AutoFill is enabled
### What happened?
Severity: High. KeeForge can terminate on both iOS and macOS after biometric unlock when Quick AutoFill is enabled and at least one database is enabled for AutoFill. In the browser credential picker flow, the picker disappears without filling the focused login field.

This affects v1.10.4 and is correlated with the multi-database identity publication changes. The issue is independent of database content.

### Steps to reproduce
1. Enable Quick AutoFill and enable at least one Database for AutoFill.
2. Launch KeeForge and unlock with Face ID on iOS or Touch ID on macOS.
3. Observe that the app terminates after biometric unlock.
4. In a browser, focus a login field and invoke AutoFill.
5. Select a credential and authenticate.
6. Observe that the credential picker disappears and the field remains empty.
7. As a workaround, disable Quick AutoFill; this avoids the app termination. Disabling Database leaves no credentials available for AutoFill.

### KeeForge version
1.10.4

### iOS version and device
Latest affected iOS on reporter device; exact version unavailable; also reproducible on macOS

### Where does the bug occur?
Main app (browsing / unlocking), AutoFill extension

Independent of database content. No real database is attached.

### Anything else?
Confirmed macOS diagnostic evidence (with local paths and private data omitted):

```
NSInvalidArgumentException: -[SFPasswordCredentialIdentity recordIdentifier]: unrecognized selector
```

The suspected root cause is the unlock flow calling `finalizeSuccessfulUnlock`, which calls `CredentialIdentityStoreManager.populate`, which enumerates `ASCredentialIdentityStore` identities. The runtime private `SFPasswordCredentialIdentity` implementation may not provide `recordIdentifier` despite the Swift API assumptions, causing the exception.

There is also a correctness issue in the AutoFill authentication flow: callbacks that indicate no user interaction must not initiate biometric UI. They should return `.userInteractionRequired` and allow the request to continue interactively.

Proposed/fixed approach:
- Use runtime-safe selector access for credential identity fields.
- Serialize process-local identity store mutations.
- Add request generation and cancellation guards.
- Guarantee exactly-once terminal completion.
- Keep logging safe and non-sensitive.
- Add focused regression tests.

Focused iOS and macOS builds/tests pass. Physical-device and system AutoFill validation requires a signed Apple Developer Program build or TestFlight build.

ftorga:
# Summary

Fixes #41

Prevent the confirmed AutoFill crash caused by runtime `SFPasswordCredentialIdentity` objects that do not expose the `recordIdentifier` Objective-C selector, and strengthen request lifecycle handling so cancellation and asynchronous unlock completion remain exactly-once.

The earlier unconditional no-user-interaction escalation has been removed from this PR. Silent Quick AutoFill remains available when it can complete safely; broader callback reliability hypotheses are tracked separately in #47.

# Changes

- Guard runtime access to credential identity record identifiers before Objective-C dispatch.
- Preserve unattributed identities when the system cannot expose enough information to remove them safely.
- Serialize process-local identity-store mutations and keep deterministic test seams.
- Invalidate cancelled or superseded unlock continuations.
- Gate terminal request completion so each request finishes exactly once.
- Preserve existing silent password, passkey, and one-time-code callback behavior.
- Ensure no-user-interaction entry points initialize request lifecycle state with `beginRequest()`.
- Restore the existing `## Unreleased` changelog entries removed during the previous merge resolution.
- Consolidate identity-store test reset/barrier helpers and replace reviewed timing assumptions with deterministic synchronization.

# Testing

- [x] `git diff --check`
- [x] `swift scripts/normalize-xcstrings.swift --check`
- [x] Focused iOS build-for-testing passed.
- [x] Focused local iOS tests passed on iPhone 17 Pro Simulator: **315 tests, 0 failures, 0 skipped**.
- [x] GitHub PR Tests passed for `4c4096e`.
- [x] DCO passed.
- [x] Internal bounded reliability audit completed with no blocking findings (this is not a GitHub approval).

Focused local test scope:

```text
KeeForgeTests/CredentialProviderCoordinatorTests        58
KeeForgeTests/CredentialIdentityStoreManagerTests       84
KeeForgeTests/CredentialProviderSaveTests               12
KeeForgeTests/DatabaseListStoreTests                    38
KeeForgeTests/DatabaseListViewModelTests                21
KeeForgeTests/DatabaseViewModelTests                   102
Total                                                  315
```

The first simulator launch failed because the simulator host process died (`NSMachErrorDomain -308` / `NSPOSIXErrorDomain Code 64`). After a targeted shutdown and reboot of that simulator, the exact focused suite passed without source changes.

## Manual validation before merge

Run on a signed physical device or TestFlight build:

- [ ] Enable KeeForge under **Settings → General → AutoFill & Passwords** and publish test identities.
- [ ] Unlock through the AutoFill extension with Face ID and confirm the original `recordIdentifier` crash does not recur.
- [ ] Cancel an unlock, retry, and confirm the request completes exactly once.
- [ ] Confirm a safe QuickType suggestion can still fill silently.
- [ ] Confirm requests that genuinely need interaction open the extension and complete normally.
- [ ] Validate password, one-time-code, and passkey paths.

# Checklist

- [x] Updated `CHANGELOG.md` under `## Unreleased`
- [x] No source files were added, removed, or moved; `xcodegen generate` was not required
- [x] Updated the nearest AutoFill documentation
- [x] No localization catalog changes were required
- [x] Parser/writer/save behavior and accessibility identifiers are unaffected


crazytan (commented):
Thanks for the thorough writeup and for splitting this into reviewable slices — the chain context made this much easier to follow.

The core crash fix is right: `recordIdentifier(of:)` guarding with `responds(to:)` before the ObjC dispatch is exactly the correct shape for `SFPasswordCredentialIdentity`, and the lifecycle/exactly-once slice is a genuine reliability win. Two things need fixing before this can merge, and I'd like to talk about the scope of slice #2.

## Blocking

**1. The merge deletes four `## Unreleased` CHANGELOG entries.** Merge-base is `3765907` (current `main` tip), so these are real deletions introduced by `c2095e2`, not a stale base — see the inline note. Four shipped fixes would disappear from the release notes.

**2. `provideCredentialWithoutUserInteraction` never calls `beginRequest()`.** Every other entry point now does; because `finishRequest` is gated on `guard !didFinishRequest`, a coordinator that already finished a request would swallow the cancel and return nothing to the system. Inline note below.

## Scope: I'd like to split slice #2 out

Issue #41 is an `unrecognized selector` crash, and the `recordIdentifier(of:)` guard fixes it. Making *every* no-UI callback escalate to `.userInteractionRequired` is an independent product decision that removes a shipped feature — every QuickType tap now opens the extension UI, and "Quick AutoFill" stops meaning "fills without opening KeeForge".

If we do take it, it needs to come with the rest of the change:

- `KeeForge/Views/SettingsView.swift` (~line 422) still tells users the copy-TOTP setting makes KeeForge "ask you to confirm ... **instead of filling silently**", and the matching CHANGELOG "Added" bullet says the same. Both now describe a silent path that no longer exists. Fixing that copy means touching `Localizable.xcstrings` **and** the `de` translation, so `LocalizationTests` needs a pass.
- `"AutoFill with KeeForge"` and `"Passkey sign-in with KeeForge"` now have zero code references but remain in both catalogs.
- `providePasskeyWithoutUserInteraction`, `provideOTCWithoutUserInteraction`, and `mustEscalateToInteractiveFill` are all left with no callers.

My suggestion: land slices #1 and #3 now — I'd merge those quickly — and open slice #2 separately so the trade-off, the settings copy, and the dead-code removal get decided together.

## Smaller notes

The lifecycle CHANGELOG entry half-overstates: "AutoFill no longer treats an internal iOS presentation dismissal as a cancelled request" — but the only change to `viewDidDisappear` is an added comment, so that behavior already existed on `main`. The second clause (late unlock can't repopulate) is real and worth keeping.

Test coverage for the new behavior is good — exactly-once completion, late-continuation invalidation, `maxConcurrentMutations == 1`, store-captured-at-invocation, and synchronous ordering. The `IdentityWithoutRecordIdentifier` fixture overriding `responds(to:)` is a neat way to simulate the system object.

One process note: the PR body says focused tests can't compile because `TagDestination` / `TagListView` / `TagEntriesView` are missing upstream. That's stale — the merge base *is* current `main`, which has the tag browser. Since the manual-validation checklist is entirely unchecked and the no-UI escalation is precisely what unit tests can't prove, could you re-run the focused slices locally and refresh that section?

Remaining comments are inline. Leaving this as a comment rather than a formal request-changes, but I'd want the two blocking items resolved before merge.

CHANGELOG.md
```diff
@@ -28,9 +28,9 @@ TODO before the first macOS release:
 - Copy an entry's verification code to the clipboard when AutoFill fills its password, for sites whose one-time code field iOS does not recognize. Off by default; turn on "Copy Verification Code on AutoFill" under Settings › AutoFill. The copy clears itself after the Clipboard Clear Timeout and never leaves the device. While the setting is on, filling a suggestion for an entry that has a verification code asks you to confirm in KeeForge rather than filling silently, because AutoFill can only write the clipboard while its panel is on screen.
 
 ### Fixed
```
crazytan, line 30: **Blocking.** This hunk replaces four existing `## Unreleased` → `### Fixed` entries instead of adding to them:

- The "Delete Entry" / "Remove Database" popover-anchoring fix
- The permanent-delete editor-wedge fix (`8b29221`)
- The iPad workspace-reset fix (`d748c5f`)
- "KeeForge now preserves unknown KDBX4 inner-header fields when saving" (`3765907`)

`git merge-base` for this PR is `3765907`, i.e. the current tip of `main` — so these are deletions made by the merge commit `c2095e2`, not an artifact of an old base. Four shipped fixes would vanish from the release notes.

Could you redo the conflict resolution so the two new AutoFill lines are appended and the existing bullets stay?

AutoFillExtension/CredentialProviderCoordinator.swift
```diff
@@ -289,19 +315,12 @@ final class CredentialProviderCoordinator {
     }
 
     func provideCredentialWithoutUserInteraction(for credentialRequest: ASCredentialRequest) {
```
crazytan, line 319: **Blocking:** this entry point (and the `ASPasswordCredentialIdentity` overload below) is the only one that doesn't call `beginRequest()`.

Since `finishRequest` opens with `guard !didFinishRequest`, a coordinator instance that already completed a request would silently swallow this cancel — the extension then returns *nothing* to the system, which is the worst AutoFill failure mode (spinner forever, no completion callback). Low probability given one-request-per-process, but it's a one-line guard against an unrecoverable state, and the asymmetry with every other entry point will bite someone later.

Separately, on the change itself: unconditionally escalating here removes the silent-fill feature rather than fixing the crash, and it orphans `mustEscalateToInteractiveFill` (line 375), `providePasskeyWithoutUserInteraction` (481), and `provideOTCWithoutUserInteraction` (1303) — see the main review comment for the scope discussion.

AutoFillExtension/CredentialProviderCoordinator.swift
```diff
@@ -513,41 +479,7 @@ final class CredentialProviderCoordinator {
     // MARK: - Passkey silent auth
 
     private func providePasskeyWithoutUserInteraction(for request: ASPasskeyCredentialRequest) {
```
crazytan, line 561: This function now has no callers — same for `provideOTCWithoutUserInteraction` (line 1303) and `mustEscalateToInteractiveFill` (line 375). `mustEscalateToInteractiveFill` also keeps its full doc comment describing the silent QuickType path that no longer exists, which will read as live behavior to the next person in this file.

If slice #2 stays, these should go with it.

AutoFillExtension/CredentialProviderCoordinator.swift
```diff
@@ -826,48 +743,115 @@ final class CredentialProviderCoordinator {
...
+                guard let self, self.isRequestActive(generation) else { return }
                 let databaseReference = try currentDatabaseReference()
-                let keyFileData = try loadAssociatedKeyFileData(for: databaseReference)
-                let compositeKey = KDBXCrypto.compositeKey(password: password, keyFileData: keyFileData)
-                try await loadEntries(
-                    compositeKey: compositeKey,
-                    databaseReference: databaseReference
+
+                #if DEBUG
+                if let unlockWorkOverride {
```
crazytan: This `#if DEBUG` / `#else` shape is duplicated four times (twice here, twice in `unlockWithBiometrics`), each repeating the full real-work call with its argument list. Easy to let the two branches drift.

Could this collapse into one seam? Something like:

```swift
private func runUnlockWork(_ work: @MainActor () async throws -> Void) async throws {
    #if DEBUG
    if let unlockWorkOverride {
        try await unlockWorkOverride()
        return
    }
    #endif
    try await work()
}
```

Then each call site is a single `try await runUnlockWork { try await self.unlockPasswordWork(...) }`.

AutoFillExtension/CredentialProviderCoordinator.swift
```diff
@@ -1021,6 +1010,8 @@ final class CredentialProviderCoordinator {
         parsedEntries = offerableEntries.filter { $0.hasPassword || $0.hasPasskey || $0.hasTOTP }
     }
 
+    private struct RequestNoLongerActive: Error {}
```
crazytan: Nit: this type declaration is wedged between `loadEntries` and `loadDatabaseData`. Worth moving up with the other nested type declarations so the parse-and-load section reads uninterrupted.

AutoFillExtension/CredentialProviderCoordinator.swift
```diff
@@ -1616,15 +1570,34 @@ final class CredentialProviderCoordinator {
...
 
     // MARK: - Error handling
 
     func cancelRequest(code: ASExtensionError.Code) {
+        finishRequest { presenter in
+            presenter.cancelRequest(withError: ASExtensionError(code))
+        }
+    }
+
+    /// Captures the presenter before cleanup and gates the terminal handoff.
```
crazytan: The comment says it "captures the presenter before cleanup", but `cleanup()` doesn't touch `presenter` — it only clears vault state — so the local capture is inert. Either drop the capture and use `presenter` directly, or reword so the comment doesn't imply a dependency that isn't there.

The gating half of the doc (exactly-once terminal handoff) is accurate and worth keeping.

KeeForge/Services/AutoFill/CredentialIdentityStoreManager.swift
```diff
@@ -311,21 +484,24 @@ enum CredentialIdentityStoreManager: Sendable {
 
             let storedIdentities = await store.credentialIdentities()
             let otherDatabaseIdentitiesPresent = storedIdentities?.contains { identity in
-                guard let recordIdentifier = identity.recordIdentifier,
+                guard let recordIdentifier = recordIdentifier(of: identity),
                       case .current(let parsed) = CredentialRecordIdentifier.parse(recordIdentifier)
                 else { return false }
                 return parsed.databaseID != databaseID
             } ?? false
+            let unattributedIdentitiesPresent = storedIdentities?.contains {
```
crazytan, line 490: Worth calling out the second-order effect here, because I don't think it's obvious from the decision-tree doc above.

When `unattributedIdentitiesPresent` is true we take the additive path — but `identitiesToRemove` in that branch filters through the *same* `recordIdentifier(of:)` helper. So on a system where enumeration returns opaque identities, nothing is ever removed; only `saveCredentialIdentities` runs. `testRefreshPreservesEnumeratedIdentityWithoutRecordIdentifier` confirms this (`fake.calls == ["saveCredentialIdentities"]`).

That's the right safety choice — better a stale suggestion than deleting an identity we can't attribute — but it means stale identities accumulate with no automatic recovery. Two suggestions:

1. `logger.warning` when this path triggers, so the condition is visible in a sysdiagnose rather than silent.
2. A line in `AutoFillExtension/README.md` noting that "Clear AutoFill Entries" is the manual escape hatch on affected systems.

KeeForge/Services/AutoFill/CredentialIdentityStoreManager.swift
```diff
@@ -275,22 +439,31 @@ enum CredentialIdentityStoreManager: Sendable {
...
-    /// affected database's next refresh; no IPC or locking is layered on top.
+    /// The queue serializes calls within this process only. The main app and
+    /// extension are separate processes and can still mutate the system store
+    /// concurrently (unlock vs. in-extension save); enumerate-then-mutate is
+    /// not atomic across that boundary. Worst case is a briefly stale or
+    /// duplicate suggestion, corrected by the affected database's next
+    /// refresh; no IPC or cross-process lock is layered on top.
     static func populate(with entries: [KPEntry], for databaseID: UUID) {
         let eligibleEntries = entries.filter { !$0.isExpired() }
+        logger.info("Starting credential identity refresh with \(eligibleEntries.count) eligible entries")
```
crazytan: These `logger.info("Starting ...")` lines (here and in `clearStore`, `removeIdentities`, `removeIdentity`) read as investigation scaffolding from debugging the crash rather than diagnostics we'd want permanently. The existing failure-path `logger.error` calls already cover what goes wrong.

I'd drop them, or keep only the one that carries real signal (the eligible-entry count) and demote the rest to `.debug`.

KeeForge/ViewModels/DatabaseViewModel.swift
```diff
@@ -1622,6 +1624,8 @@ final class DatabaseViewModel {
         synchronizeSelections()
         startInactivityTimer()
 
+        Self.logger.info("Database unlock finalized; starting shared credential identity refresh")
```
crazytan: This adds an `OSLog` import and a whole `Logger` to `DatabaseViewModel` for a single `info` line on the unlock path. Same read as the manager-side logging — leftover from tracking down the crash rather than something we'd keep. Suggest removing both this and the `logger` property.

KeeForgeTests/CredentialProviderCoordinatorTests.swift
```diff
@@ -33,14 +37,19 @@ final class CredentialProviderCoordinatorTests: XCTestCase {
         let sharedDefaults = UserDefaults(suiteName: SharedVaultStore.appGroupID) ?? .standard
         sharedDefaults.removeObject(forKey: "KeeForge.quickAutoFillEnabled")
         sharedDefaults.removeObject(forKey: "KeeForge.autoFillCopyTOTP")
+        sharedDefaults.removeObject(forKey: "KeeForge.autoUnlockWithFaceID")
     }
 
     private func resetCredentialIdentityStoreSeams() {
+        resetCredentialIdentityObservers()
```
crazytan: The `waitForPendingMutations()` + split `resetCredentialIdentityObservers()` pattern is now copy-pasted verbatim into six test files (`CredentialIdentityStoreManagerTests`, this one, `CredentialProviderSaveTests`, `DatabaseListStoreTests`, `DatabaseListViewModelTests`, `DatabaseViewModelTests`).

Since the queue is process-wide, the barrier discipline is a shared invariant — worth one `XCTestCase` extension in `KeeForgeTests/` so a future fix lands in one place instead of six.

KeeForgeTests/CredentialProviderCoordinatorTests.swift
```diff
@@ -148,6 +157,54 @@ final class CredentialProviderCoordinatorTests: XCTestCase {
...
+        }
+
+        coordinator.prepareCredentialList(for: [githubServiceIdentifier()])
+        coordinator.presentUnlockPromptIfNeeded()
+        let prompt = try XCTUnwrap(presenter.unlockPrompt)
+        prompt.onChooseBiometrics()
+
+        await fulfillment(of: [unlockEntered], timeout: 1)
+        coordinator.cancelRequest(code: .userCanceled)
+        try await Task.sleep(for: .milliseconds(100))
```
crazytan: This asserts the *absence* of a late presentation after a fixed 100ms sleep, which makes the test's correctness a timing assumption. The RC workflow runs on an iOS 18 iPhone SE (3rd gen) simulator, and that's where I'd expect this to flake first.

Could the invalidation be observed deterministically instead — e.g. have `unlockWorkOverride` signal after it resumes from cancellation, and await that, rather than sleeping past it?

Related: `testUnlockRefreshesOnlyTheUnlockedDatabase` now asserts `fake.calls == ["saveCredentialIdentities"]` immediately after the first `onMutation` fires. An exact-equality assertion on a list that a second mutation could still append to would be safer behind `waitForPendingMutations()`.

CHANGELOG.md
ftorga: Fixed in c4b5db8. The four existing Unreleased entries are restored and the AutoFill entries are appended.

AutoFillExtension/CredentialProviderCoordinator.swift
ftorga: Fixed in c4b5db8. Both no-user-interaction entry points now call beginRequest(), and the unconditional escalation was removed so safe silent completion remains available.

AutoFillExtension/CredentialProviderCoordinator.swift
ftorga: Fixed in c4b5db8. Silent password, passkey, and OTP paths are restored, and mustEscalateToInteractiveFill has live callers again.

AutoFillExtension/CredentialProviderCoordinator.swift
ftorga: Fixed in c4b5db8. Password and biometric unlock now share one runUnlockWork seam, avoiding duplicated DEBUG and production call shapes.

AutoFillExtension/CredentialProviderCoordinator.swift
ftorga: Fixed in c4b5db8. RequestNoLongerActive now lives with the coordinator nested declarations rather than interrupting the loading methods.

AutoFillExtension/CredentialProviderCoordinator.swift
ftorga: Fixed in 4c4096e. The comment now describes exactly-once terminal gating and presenter capture for the completion action without implying that cleanup clears presenter.

KeeForge/Services/AutoFill/CredentialIdentityStoreManager.swift
ftorga: Fixed in c4b5db8. The opaque-identity additive path emits a warning, and AutoFillExtension/README.md documents Clear AutoFill Entries as the manual recovery path.

KeeForge/Services/AutoFill/CredentialIdentityStoreManager.swift
ftorga: Fixed in c4b5db8. The investigation-only Starting logs were removed; the remaining logging carries outcome or failure signal.

KeeForge/ViewModels/DatabaseViewModel.swift
ftorga: Fixed in c4b5db8. The DatabaseViewModel OSLog import, logger property, and unlock-path info log were removed.

KeeForgeTests/CredentialProviderCoordinatorTests.swift
ftorga: Fixed in 4c4096e. A single XCTestCase helper now drains process-wide mutations before and after resetting all identity-store test seams, and all six suites use that shared invariant.

KeeForgeTests/CredentialProviderCoordinatorTests.swift
ftorga: Fixed in c4b5db8. Cancellation now waits for an explicit resumed-after-cancellation signal instead of sleeping, and the exact call-list assertion is protected by the pending-mutation barrier.

CHANGELOG.md
```diff
@@ -25,12 +25,16 @@ TODO before the first macOS release:
...
-- Copy an entry's verification code to the clipboard when AutoFill fills its password, for sites whose one-time code field iOS does not recognize. Off by default; turn on "Copy Verification Code on AutoFill" under Settings › AutoFill. The copy clears itself after the Clipboard Clear Timeout and never leaves the device. While the setting is on, filling a suggestion for an entry that has a verification code asks you to confirm in KeeForge rather than filling silently, because AutoFill can only write the clipboard while its panel is on screen.
+- Copy an entry's verification code to the clipboard when AutoFill fills its password, for sites whose one-time code field iOS does not recognize. Off by default; turn on "Copy Verification Code on AutoFill" under Settings › AutoFill. The copy clears itself after the Clipboard Clear Timeout and never leaves the device.
```
crazytan, line 28: This trim goes the wrong way now that silent fill is restored: the deleted sentence — "While the setting is on, filling a suggestion for an entry that has a verification code asks you to confirm in KeeForge rather than filling silently…" — describes behavior that still exists. `provideCredentialWithoutUserInteraction` still routes through `mustEscalateToInteractiveFill` (CredentialProviderCoordinator.swift:413/425), so a copy-TOTP user will see exactly this confirmation bounce. Please restore the sentence; it was accurate on `main` and is accurate again on this head.

CHANGELOG.md
```diff
@@ -25,12 +25,16 @@ TODO before the first macOS release:
...
 - Permanently deleting an entry from the entry editor ("Delete Entry" → "Delete Permanently") left the app stuck on the editor screen until force-quit: deleting the entry tore down the navigation destination that had presented the editor, so nothing could pop it. The deletion itself was saved correctly. The editor now dismisses and returns to the group list, including when the entry was already in the recycle bin. ("Move to Recycle Bin" was unaffected.)
 - On iPad, that same permanent delete also abruptly reset the whole workspace: the sidebar snapped back to the vault root mid-save and the editor vanished without animation. The detail screen's close was popping the sidebar's navigation stack (its own column had nothing to pop), and the editor was torn down before the save finished. Permanent delete now unwinds exactly like "Move to Recycle Bin": the editor pops after the save completes and the sidebar stays on the open group.
+- KeeForge now preserves unknown KDBX4 inner-header fields when saving.
```
crazytan, line 35: Nit: the restored entry lost its second clause. On `main` it reads "…when saving, protecting data written by future KeePass format extensions." Worth restoring verbatim so the merge doesn't silently edit an entry this PR isn't about.

AutoFillExtension/README.md
```diff
@@ -19,11 +19,13 @@ This target provides password, passkey, one-time-code, and new-credential save/g
...
+- Opt-in "Copy Verification Code on AutoFill" (`SettingsService.autoFillCopyTOTP`, App-Group-shared, default off): when a filled password entry also has a TOTP config, `completeRequest(with:)` generates the current code and copies it via the shared `../KeeForge/Services/AppSupport/ClipboardService.swift` before `cleanup()` drops the session key. The copy goes through the coordinator's injectable `copyToClipboard` seam and inherits the App-Group-shared Clipboard Clear Timeout as a `UIPasteboard` expiration date, so it expires after the extension is gone. iOS only (`#if os(iOS)`): the macOS `ClipboardService` branch clears via an in-process timer that dies with the extension. The toggle lives in Settings → AutoFill (identifier `settings.autofill.copy-totp`; recently moved from the Security section).
```
crazytan, line 28: Same issue as the CHANGELOG: this bullet's closing sentence on `main` — "Because a no-UI extension cannot reliably write the pasteboard, `provideCredentialWithoutUserInteraction` escalates to `.userInteractionRequired` instead of filling silently whenever the setting is on and the chosen entry has a verification code." — was replaced in the previous revision and then dropped entirely in `c4b5db8`. With the silent path restored, the original `main` wording is correct again and this is the only doc that explains *why* the escalation happens. Please restore it.

crazytan (commented):
Re-reviewed `c4b5db8` and `4c4096e`. Both blocking items are resolved, and every inline thread was addressed as described — thank you for the fast, precise turnaround.

Restoring the silent paths with the generation guards woven in is a better resolution than either option I suggested: the crash fix lands, the feature stays, and the no-UI paths now get the same `isRequestActive` / exactly-once discipline as interactive unlock. I walked the three restored paths (password, passkey, OTC) against the lifecycle machinery — generation captured after `beginRequest()`, `RequestNoLongerActive` thrown from `loadEntries` and swallowed when stale, `defer` blocks generation-checked so `finishRequest`'s own bump isn't clobbered — and found no gaps. The dispatcher calling `beginRequest()` before handlers that call it again double-increments the generation, but it's captured after the second bump, so it's harmless.

The test changes are also strictly better: the shared `resetCredentialIdentityStoreState()` barrier, the deterministic cancellation signal replacing the 100ms sleep, and the remaining `Task.sleep` negatives converted to `waitForPendingMutations()`.

One regression introduced by the revert, plus a nit — both inline. The revert restored the escalate-when-copy-TOTP-is-on behavior (`mustEscalateToInteractiveFill` has live callers again at lines 413/425), but `c4b5db8` also *deleted* the CHANGELOG and README sentences documenting that behavior. Those sentences are true again and should come back. (The `SettingsView` footer, which still describes it, is now accurate as-is — no change needed there.)

With those doc restorations this is good to merge from my side. The manual on-device validation checklist in the PR body is still the real gate — the no-UI → `.userInteractionRequired` → interactive-relaunch transition is exactly what unit tests can't prove.

PR #45, opened 2026-07-27
feat(entries): add a read-only entry history viewer

miquno:
# Summary

- Split as requested in #38 (closed; this replaces it so the thread matches the branch). This is the read-only half — it surfaces the versions KDBX already stores; restoring is #44.
- An entry shows a History row that opens its stored `<History>` versions, and one version can be inspected read-only.

# Changes

- Add `EntryHistoryView`: version list plus a read-only version screen. A sheet, since pushed levels inside the macOS sidebar column render zero-height.
- Add `DatabaseViewModel.history(forEntryID:)`, sorting versions newest-first.
- Add an `entry-detail.history` row, shown only when versions exist.
- Give `PasswordFieldRow` an `accessibilityPrefix` (default `entry`) so the history screen's reveal/copy ids do not collide with entry detail's.
- Localize the new strings in `en` and `de`; document the screen in `KeeForge/Views/README.md`.

Sorting is deliberate rather than trusting the file: KDBX fixes no order for `<History>`, and the conventions disagree — KeePass and KeePassXC append chronologically while this app's edit path prepends. Verified against a database authored by `keepassxc-cli`, whose oldest version arrives at index 0; without the sort, every database KeeForge did not write listed its oldest version on top.

Nothing here writes: no `EntryEdit` case, no `DatabaseDraft` change, no compatibility-matrix change.

# Testing

- [x] `xcodebuild test … -only-testing:KeeForgeTests/{DatabaseViewModelTests,LocalizationTests}` — 112 tests passed, 0 failed.
- [x] `xcodebuild test … -only-testing:KeeForgeUITests/EntryHistoryUITests` — 2 tests passed, 0 failed.
- [x] Read a database whose `<History>` was authored by `keepassxc-cli` (KeePassXC 2.7.12): both stored versions come back, including their decrypted passwords.
- [x] Walked by hand on an iPhone 17 Pro and an iPad Pro 11-inch simulator against the `test` fixture.

# Checklist

- [x] Updated `CHANGELOG.md` under `## Unreleased`.
- [x] Ran `xcodegen generate` after adding `EntryHistoryView.swift`.
- [x] Updated `KeeForge/Views/README.md`.
- [x] Localized new strings in `en` and `de`, ran `normalize-xcstrings.swift` and `LocalizationTests`.
- [x] Parser, writer, save behavior and the compatibility matrix are unaffected — this PR adds no edit operation.
- [x] AutoFill target membership unchanged; the new view is app-only.
- [x] New identifiers added, existing ones unchanged (`PasswordFieldRow` keeps its `entry.*` defaults).
- [x] Locking clears the history with the rest of the session (`testHistoryIsUnreachableAfterLocking`), so an open viewer cannot outlive the session key.

Known gaps:

- No "what changed" diff between versions; the list shows date and title only.
- macOS is untouched by design — the targets are on hold, and the sheet was chosen so nothing depends on sidebar pushes.


crazytan (commented):
Thanks for splitting this out — the scope discipline is right, and the two things that could have gone wrong are both handled: history entries are sealed with the session key at parse time (`KDBXParser.swift:1246` builds them via `builder.build(sessionKey:)`), and `entryIndex` is built from live entries only, so the UUID that history versions share with their parent can't alias the index. `testHistoryIsUnreachableAfterLocking` is exactly the test I'd have asked for.

I checked out the branch and re-ran `-only-testing:KeeForgeTests/DatabaseViewModelTests -only-testing:KeeForgeTests/LocalizationTests` — 85 tests, all four history tests pass. (One run reported a failure in `testSaveAsConflictCopyLocalWritesSiblingFileClearsConflict`, but the log shows the test host crashed and the test was never re-run; it passes in isolation on this branch. Unrelated to this PR.)

Three things I'd like fixed before merge, plus one adjacent bug this PR exposes.

---

### 1. "Replaced" mislabels the timestamp

`KeeForge/Views/EntryHistoryView.swift:100` labels `version.lastModificationTime` as **Replaced** (de "Ersetzt am"), and the list row presents the same value the same way.

That timestamp is when the version was *last edited*, not when it stopped being current. `cloneForHistory()` (`Entry.swift:222`) copies the original entry as-is, keeping its `lastModificationTime`, and `makeUpdatedEntry` then stamps the *new* live entry with `timestamp` (`DatabaseDraft.swift:456`). KeePass does the same — `CreateBackup()` runs before `Touch()`. So a version's timestamp is when that version came into being; it was replaced at the *next* version's timestamp.

Concretely: entry created 10:00, edited 12:00 → the stored version reads "Replaced 10:00", but it was replaced at 12:00. For a feature whose pitch is "check what a password used to be, and when", that's off by one version.

KeePassXC labels this column "Last modified". Suggest matching that — "Last Modified" / "Zuletzt geändert".

### 2. The German string is informal; the rest of the catalog is formal

> "Earlier versions appear here after you edit this entry." → "…sobald **du** diesen Eintrag bearbeitest."

Every other German string in `Localizable.xcstrings` uses *Sie* — I counted 99 Sie-form strings, and this is the only du-form one. Should be:

> "Frühere Versionen erscheinen hier, sobald **Sie** diesen Eintrag bearbeiten."

`LocalizationTests` won't catch this — it checks for missing/drifted translations and format-specifier parity, not register.

### 3. The documented `entry-history.{title,username,url}` identifiers don't exist

`FieldRow` hardcodes `accessibilityID: "entry.copy.\(normalizedLabel)"` (`EntryDetailView.swift:410`), so passing `accessibilityKey: "entry-history.title"` actually produces **`entry.copy.entry-history.title`**. The `Views/README.md` addition claims `entry-history.{title,username,url}`, so a UI test written from the docs would fail.

It does avoid collisions — the namespace is smuggled into the key — but it's inconsistent with the `accessibilityPrefix` treatment you gave `PasswordFieldRow` immediately below it. Either give `FieldRow` the same `accessibilityPrefix` parameter, or correct the README to the real identifiers.

---

### Adjacent: this PR's own rationale exposes a history-trimming bug (out of scope, worth a follow-up issue)

Your reasoning — KeePass/KeePassXC append oldest-first, KeeForge prepends — means `trimmedHistory` is trimming from the wrong end. `DatabaseDraft.swift:563`:

```swift
var history = ([snapshot] + existing).map { $0.cloneForHistory() }
if maxItems >= 0, history.count > maxItems {
    history = Array(history.prefix(maxItems))
}
```

For a KeePass-authored database `existing` is oldest-first, so `[v4, v1, v2, v3].prefix(3)` keeps `v4, v1, v2` and drops **v3, the second-newest**. Once an entry reaches the default 10-version cap, every subsequent save silently discards recent history and retains the oldest. The `maxSize` loop below has the same assumption.

Pre-existing, not introduced here — but this viewer is what will make it visible to users, so it seems worth filing separately.

---

### Smaller

- The UI test asserts on the localized visible string `app.staticTexts["Earlier Version"]`. `Views/README.md` explicitly says to add identifiers rather than rely on visible labels — worth one on the version screen.
- `EntryHistoryUITests` isn't listed in `KeeForgeUITests/README.md`'s class catalog, nor in the "files hosting multiple classes" line for `UnlockedDatabaseUITests.swift`. (`GroupIconPickerUITests` is missing too, so it's not only you.)
- No tiebreak for equal timestamps in `history(forEntryID:)`. KDBX timestamps are 1-second resolution, `versions` is recomputed independently by the list and by `EntryHistoryVersionView`, and `Array.sorted` isn't documented as stable — so in principle the row you tap isn't the version you get. You already added an `offset` tiebreak for `(nil, nil)`; extending it to the equal-date case makes the order a total one.
- `String(localized: "(untitled)")` has no catalog entry (nor do its three pre-existing literal uses), so German falls back to English there.
- The version screen omits custom fields, tags, TOTP, expiry and creation time — a stored version can differ in all of them. Custom fields seem the likeliest to matter; worth adding to "Known gaps" at minimum.
- Notes render as plain `Text` here vs. `SelectableNotesText` in entry detail (no selection/sizing parity).
- `Text("\(entry.history.count)")` isn't locale-formatted; `Text(entry.history.count, format: .number)` is.

Issue #46, opened 2026-07-27
feat(autofill): create a new entry from the AutoFill picker


miquno:
### What problem would this solve?

Signing up for something new is the one moment where KeeForge makes me leave the keyboard.

Picture the usual signup page. I tap the password field, KeeForge's suggestions come up, and there is nothing there yet — of course, the account does not exist. At that point I want to create it right where I am. Instead I have to leave the page, switch to KeeForge, make the entry, copy the password, switch back and paste it in.

The odd part is that the creator screen already exists. It just cannot be opened on purpose. It only appears when iOS decides to ask "save this password?", which happens *after* the form is submitted — too late to help me pick a password in the first place.

### Proposed solution

Let the credential picker create an entry, instead of only choosing an existing one.

Add a way into the creator from the picker, and have it arrive already filled in: name and URL taken from the site I am on, and a generate button next to the password field so I never have to think one up. Once saved, fill it into the form and publish it, so the entry is waiting in the QuickType bar the next time I come back.

iOS only — `ASSavePasswordRequest` does not exist on macOS, and the creator is already gated that way.

### Alternatives considered

Bitwarden solves it the way I would expect: its picker has a floating **+**, and the screen it opens already knows the domain — it lands in both the item name and the website field — with a generate button sitting right next to the password. Nothing to type, nothing to think about.

The alternative is the app-switching dance above, which is exactly what I would like to stop doing.

### Anything else?

Looking at the code, this seems like connecting things that are already there rather than building something new:

- `AutoFillSaveCoordinator.initialDraft(for:username:password:)` already derives name and URL from the service identifier — it is what the save-password path uses.
- `PasswordGenerator` is already compiled into the extension and already serves `ASGeneratePasswordsRequest`.
- `AutoFillEntryCreatorView` is the form itself.

What is missing is a route to the creator outside the save-password request, and a generate control on its password field.

Happy to build it if you want it — asking first, since it changes what the extension shows. Two things I would want your call on:

1. Floating **+** like Bitwarden, or a plain toolbar button next to Cancel?
2. Where should a new entry land? The creator has no group picker today, so either it goes somewhere fixed or it needs one.

crazytan:
@miquno Interesting idea! Do you know if either KeePassium or Strongbox does this, Or this is a feature from Bitwarden only? KeePassXC creates a default group for saving entries created from browser extension so maybe we can use similar patterns.

miquno:
[...]

PR #75, opened 2026-08-03
fix(autofill): let the picker show every credential

miquno:
# Summary

The picker is a dead end when matching finds nothing: the coordinator has pre-filled the search with the domain it just failed to match, so the list renders empty and reports "No Credentials Found". The whole fillable corpus is already in `searchEntries`, but reaching it means clearing a search field the user never typed into. Adds "Show All Credentials".

# Changes

- `canShowAllEntries` in `AutoFillSearchView` detects the gap (`filteredEntries + possibleEntries < searchEntries`); taking the action clears the search and sets `showsAllEntries`, so the unfiltered branch returns `searchEntries` instead of the narrowed `entries`.
- Offered in two places: the `ContentUnavailableView` empty state, and a row below a short list of matches.
- Not offered on pickers whose `entries` already are `searchEntries` (passkey, single-entry), nor once the user edits the search themselves.
- View-local state only — no coordinator change.
- Localized "Show All Credentials" in `de`, `fr`, `es`; `AutoFillExtension/README.md` and `CHANGELOG.md` updated.

# Testing

- `xcodebuild test -project KeeForge.xcodeproj -scheme KeeForge -destination 'platform=iOS Simulator,name=iPhone 17 Pro' -only-testing:KeeForgeTests` — 1373 passed, 0 failed, 0 skipped (Xcode 26.6, iOS 26.5).
- `xcodebuild build -scheme KeeForgeMac -destination 'platform=macOS'` — succeeded; the mac shell hosts the same view.
- **No unit test for the behavior itself.** It is SwiftUI `@State` that the coordinator suites cannot reach through `CredentialProviderPresentingSpy`, and no placeholder test was added to paper over that.
- Verified by hand in the extension instead, on the `KeeForge-AutoFill-Harness` simulator (`scripts/provision-autofill-harness-sim.sh`) with `TestFixtures/demo.kdbx` (16 entries) and a local login form on an unmatched host: the picker opened on the coordinator's pre-filled `localhost` search reporting "No Credentials Found", the empty state offered "Show All Credentials", and taking it cleared the search and listed the full set (Twitter/X, Discord, Reddit, Mastodon, Chase Bank, Coinbase, Robinhood, GitHub, AWS Console, Jira, Slack, …).

# Checklist

- [x] Updated `CHANGELOG.md` under `## Unreleased` (`### Fixes`)
- [x] Ran `xcodegen generate` — no source files added/removed, so the project file is unchanged
- [ ] Updated `KDBXCompatibilityTests` if parser/writer/save behavior changed — n/a, no parser/writer/save change
- [x] Preserved accessibility identifiers or updated the affected UI tests — existing identifiers untouched; new `autofill.show-all-entries` and `autofill.show-all-entries.empty-state` added

# Notes

- Pre-existing and untouched: with the full list shown, entries already listed under "Possible matches" appear twice. The `didEditSearch` path does the same today. Hiding that section would remove the only route to "Add original URL to this entry".
- Independent of #46 / `feat/autofill-picker-create-entry`; both branch from `main`. Whichever lands second needs a small rebase in `AutoFillSearchView.swift` (both add to the same `ContentUnavailableView` `actions:` block) and in `CHANGELOG.md`. The resolution is verified — the two features compile and test green together.


AutoFillExtension/AutoFillSearchView.swift
```diff
@@ -56,6 +57,21 @@ struct AutoFillSearchView: View {
...
     }
 
+    /// Whether the picker is narrowed to matches — or to a domain search the
+    /// coordinator pre-filled — while further credentials sit behind it. A
+    /// request whose matching found nothing lands on that pre-filled search
+    /// with an empty list, so without this the full database is reachable
+    /// only by clearing a search field the user never typed into.
+    private var canShowAllEntries: Bool {
+        guard !showsAllEntries, !didEditSearch, !searchEntries.isEmpty else { return false }
+        return filteredEntries.count + possibleEntries.count < searchEntries.count
```
crazytan, line 67: The verification-code picker still has a branch where this can never become true. When `presentOTCMatchesOrFinish()` has multiple matches, it calls `presentSearchView(entries: matches, ...)` without passing `allTOTPEntries` as `searchEntries`; the helper therefore makes `searchEntries == entries`, so unrelated TOTP credentials remain unreachable and Show All is hidden. Could that call pass `searchEntries: allTOTPEntries` and add coverage for multiple matches plus an unrelated TOTP entry?

miquno: Good catch — confirmed and fixed in aae3b9d.

The multi-match branch passed only `entries: matches` and let `searchEntries` default to the same array, so `canShowAllEntries` compared a set against itself and could never be true. The verification-code claim in the CHANGELOG was therefore only true for the no-matches branch. It now passes `searchEntries: allTOTPEntries`, mirroring what the password path already does with `allPasswordEntries`.

Covered by `test_otcList_multipleMatches_keepsUnrelatedTOTPEntriesInSearchCorpus`: two `github.com` TOTP entries plus an unrelated one, asserting the listed set stays the two matches while the search corpus keeps all three. Verified it fails with the fix reverted.

PR #76, opened 2026-08-03
feat(autofill): create a new entry from the AutoFill picker

### LINKED ISSUE #46: feat(autofill): create a new entry from the AutoFill picker

miquno:
# Summary

Signing up is the one moment the picker has nothing to offer: the account does not exist yet, so the list is empty and the way forward is to leave the page, switch to KeeForge, make the entry, copy the password and come back. The creator screen already exists — it just cannot be opened on purpose. The picker now opens it directly, prefilled from the request, and fills the form with what was saved.

Fixes #46.

# Changes

- `CredentialProviderCoordinator.makeEntryCreationAction()` hands the picker an `onCreateEntry` only when the request can actually produce an entry: iOS, writable database, not KDBX 3.1, and a service identifier present. Passkey and one-time-code pickers stay selection-only; the mac shell always receives `nil`.
- `AutoFillSearchView` shows a toolbar `+` and, in the empty state, a "Create New Credential" button — both only when `onCreateEntry` is non-nil.
- `presentEntryCreator(for: ASCredentialServiceIdentifier)` reuses `AutoFillSaveCoordinator.initialDraft(for:username:)` with an empty username and a generated password.
- `saveNewEntryAndFill` saves through `AutoFillSaveCoordinator.saveNewEntry`, then completes with an `ASPasswordCredential` built from the confirmed draft — not with `completeSavePasswordRequest`, which would leave the form unfilled. The credential comes from the draft rather than from locating the saved entry: the plaintext is in hand and a new entry has no TOTP config to copy. QuickType publication is inherited from `saveNewEntry`.
- `AutoFillEntryCreatorView.allowsPasswordEditing` — `false` for save-password requests, whose password the site already received; `true` here, which unlocks the field and adds a generate button via the shared `PasswordInputRow` (reveal comes with it).
- iOS shell dismisses the picker before presenting the creator, same handling as `onSelect`/`onCancel`.
- New `KeeForgeTests/CredentialProviderPickerCreationTests.swift`; `CredentialProviderPresentingSpy` now records the creator presentation.
- 3 new strings localized in `de`, `fr`, `es`; both folder-local `README.md`s and `CHANGELOG.md` updated; `xcodegen generate` run.

# Testing

- `xcodebuild test -project KeeForge.xcodeproj -scheme KeeForge -destination 'platform=iOS Simulator,name=iPhone 17 Pro' -only-testing:KeeForgeTests` — 1381 passed, 0 failed, 0 skipped (Xcode 26.6, iOS 26.5). 8 of those are the new `CredentialProviderPickerCreationTests`: when the action is offered at all (writable / KDBX 4 / iOS / service identifier present), what the creator is prefilled with, the cancel path, and the two save refusals (no title or username, missing vault state). The successful save-and-fill is **not** unit-tested — it needs a real vault and a real save.
- `xcodebuild build -scheme KeeForgeMac -destination 'platform=macOS'` — succeeded; the mac shell's `CredentialProviderPresenting` conformance follows the widened protocol.
- Exercised end to end in the extension on the `KeeForge-AutoFill-Harness` simulator (`scripts/provision-autofill-harness-sim.sh`) with `TestFixtures/demo.kdbx` and a local login form on an unmatched host: picker showed the toolbar `+` and the empty-state "Create New Credential"; the creator opened prefilled (title `localhost`, empty username, generated password, request URL); "Save and Fill" wrote the entry and filled the form with `testuser` + the generated password; reopening the database showed the new `localhost / testuser` entry.
- That manual pass is what caught the generate-button defect below — the unit suite was green with the button missing.

# Checklist

- [x] Updated `CHANGELOG.md` under `## Unreleased` (`### New Features`)
- [ ] Updated `KDBXCompatibilityTests` if parser/writer/save behavior changed — n/a, this reuses the existing `saveNewEntry` write path unchanged
- [x] Preserved accessibility identifiers or updated the affected UI tests — existing identifiers untouched; new `autofill.create-entry`, `autofill.create-entry.empty-state`, `autofill-entry-creator.password-visibility`, `autofill-entry-creator.password-generator-button` added

# Notes

- `PasswordInputRow`'s `actions` is passed labelled, not as a trailing closure. `onVisibilityToggle` precedes it and also takes a closure, so a trailing one binds *there* — the generate button then never renders and the compiler only says "result of call to 'accessibilityIdentifier' is unused". `EntryEditView` avoids this by passing `onVisibilityToggle:` explicitly; this call site has no toggle callback, so it names `actions:` instead.
- Design decisions follow what was proposed in #46 and are still awaiting @crazytan's word: toolbar `+` rather than a floating button (a floating button fights the `.searchable` field and the keyboard), and new entries into `rootGroup.groups.first` — the database's top-level group, the same place the save-password path already writes — rather than a dedicated AutoFill group. Writing into the synthetic KDBX Root is what corrupts databases for other KeePass apps (keepassium/KeePassium#476); KeePassXC's auto-created group drew pushback (keepassxreboot/keepassxc#3119). A configurable group can come later.
- Independent of `fix/autofill-picker-show-all`; both branch from `main`. Whichever lands second needs a small rebase in `AutoFillSearchView.swift` (both add to the same `ContentUnavailableView` `actions:` block) and in `CHANGELOG.md`. The resolution is verified — the two features compile and test green together.


AutoFillExtension/CredentialProviderCoordinator.swift
```diff
@@ -1592,8 +1606,103 @@ final class CredentialProviderCoordinator {
...
+
+    /// Saves the entry, then fills the form with it. The credential is built
+    /// from the draft the user just confirmed rather than by locating the
+    /// saved entry: the plaintext is already in hand, and a brand-new entry
+    /// has no TOTP config for `completeRequest(with:)` to copy.
+    private func saveNewEntryAndFill(
+        draftPayload: EntryDraftPayload
+    ) async -> CredentialProviderEntrySaveOutcome {
+        let user = draftPayload.username.isEmpty ? draftPayload.title : draftPayload.username
+        guard !user.isEmpty else {
```
crazytan, line 1665: Because this flow makes the password editable, the user can clear the generated value and still save. This guard only validates the user value, so `saveNewEntryAndFill` then persists a passwordless entry and completes with an empty `ASPasswordCredential`; the saved entry is also omitted from password identity publication because `hasPassword` is false. Could this reject an empty password (or disable Save and Fill while it is empty) and cover that case in the picker-creation tests?

miquno: Right on both counts — fixed in 384419f.

Added a guard that rejects an empty password before anything is written, so the entry cannot be persisted in a state `hasPassword` rejects and the request cannot complete with an empty `ASPasswordCredential`. New string localized in de/fr/es.

Covered by `test_saveFromCreator_emptyPassword_showsErrorWithoutSavingOrFilling`. Worth flagging: my first version of that test was worthless. It asserted only `guard case .showError`, and the seeded test vault fails the save anyway, so it passed with the guard removed — the sibling `missingTitleAndUsername` test I wrote earlier had the same hole. Both now assert the specific message, and I verified each fails when its guard is reverted.

PR #78, opened 2026-08-04
feat(entries): choose an entry's icon

miquno:
# Summary

Entries carried an `<IconID>` the format supports and no way to change it: the standard icon set was reachable for groups only, and a custom icon — the images a database keeps in `Meta/CustomIcons`, which is where other clients put a downloaded favicon — could be displayed but never picked.

Tapping the icon in the entry detail header now opens a picker with both, split into two grids the way KeePass's own chooser splits them. The two are exclusive because the format makes them exclusive: `<CustomIconUUID>` outranks `<IconID>` in every client, so a standard pick clears the custom icon rather than sitting alongside it, and the sheet marks whichever one is actually on screen.

`setEntryIcon` is the first edit that *writes* a `<CustomIconUUID>`. The serializer emits that element solely from the preserved XML, so the selection has to land there or it is lost on the next save: the edit drops any existing element and appends one at the position the writer reaches directly after `<IconID>`, which is where the schema puts it.

That position is deliberately **not** a constant. A parsable `<Binary>` advances the opaque-XML position space while leaving the anchor `KPAttachment.insertionIndex` is expressed against alone (`KDBXParser`'s `case "Binary" where currentBinaryRefWasParsable`), so an entry whose source interleaved attachments before `<IconID>` shifts every later fragment; `customIconUUIDInsertionIndex(for:)` counts those slots. A fixed 2 writes the element ahead of `<IconID>` on exactly those entries — `test_setEntryIcon_customLandsAfterIconIDEvenWithAnEarlyAttachment` pins it.

UUID-to-base64 moves into a single `KDBXUUIDEncoding` so the fragment and the serializer cannot drift on a detail the format does not forgive.

**First of two slices.** Adding an image to `Meta/CustomIcons` is deliberately out of scope here — that section is round-tripped verbatim, so this picker can only offer icons the database already carries. Downloading a favicon into the database is the follow-up and needs the writer opened up for it, which is a change to stable-core files and wants its own review.

# Changes

- `KeeForge/Views/EntryIconPickerView.swift` — new: two grids, custom section omitted when the database defines none, unrenderable icon data skipped rather than drawn as an empty cell, scrolls to the icon in use on appear.
- `KeeForge/Views/EntryDetailView.swift` — the header icon becomes the entry point. Read-only databases render it as a plain image with no button, rather than offering a chooser that could only report a refusal.
- `KeeForge/Models/EntryEdit.swift` + `DatabaseDraft.swift` — `EntryIconSelection` and `setEntryIcon`, which pushes a `<History>` version like every other entry edit.
- `KeeForge/ViewModels/DatabaseViewModel.swift` — `setEntryIcon(_:entryID:)` with the same screening `setGroupIcon` does, plus `customIcons` sorted by UUID (KDBX fixes no order for that dictionary).
- `KeeForge/Models/KDBXUUIDEncoding.swift` — new, and `KDBXXMLSerializer.serializeUUID` now calls it. No behavior change; `KDBXRoundTripTests` pins that.
- CHANGELOG under `## Unreleased`, both folder READMEs, and the new strings in `en`, `de`, `fr`, `es`.

# Testing

```
xcodebuild test -project KeeForge.xcodeproj -scheme KeeForge \
  -destination 'platform=iOS Simulator,name=iPhone 17 Pro' \
  -only-testing:KeeForgeTests/DatabaseDraftTests \
  -only-testing:KeeForgeTests/DatabaseViewModelTests \
  -only-testing:KeeForgeTests/KDBXRoundTripTests \
  -only-testing:KeeForgeTests/KDBXCompatibilityTests \
  -only-testing:KeeForgeTests/LocalizationTests
```

- 279 tests, 0 failures. New draft tests cover: a standard pick clearing the element, a custom pick landing directly after `<IconID>` in the serialized XML (including with an attachment interleaved before it), a second custom pick replacing rather than duplicating it, and the history version keeping the icon it was saved with. New view-model tests cover the screening (index outside the standard set, a custom icon the database does not define, and re-picking the current icon staying a no-op).
- `-only-testing:KeeForgeUITests/EntryIconPickerUITests` — 2 tests, 0 failures: the choice round-trips through the draft and back into the reopened picker, and cancelling changes nothing.
- `xcodebuild build -scheme KeeForgeMac -destination 'platform=macOS' CODE_SIGNING_ALLOWED=NO` — succeeds.

# Checklist

- [x] Kept AutoFill extension target membership/imports in sync for shared code — both new files land where the existing allow-list rules already place them (`KeeForge/Models` wholesale, `KeeForge/Views` app-only), so no allow-list edit was needed; `KDBXUUIDEncoding` is Foundation-only and extension-safe
- [x] Updated `KDBXCompatibilityTests` if parser/writer/save behavior changed — the supported matrix is unchanged (no new element is written for a database that had none); the suite was run as a regression check on the serializer refactor
- [x] Preserved accessibility identifiers or updated the affected UI tests — new identifiers added with UI tests covering them


KeeForge/Resources/Localizable.xcstrings
```diff
@@ -3826,6 +3826,50 @@
...
+        "es" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "Icono personalizado"
+          }
+        },
+        "fr" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "Icône personalizada"
```
crazytan: The French translation for `"Custom icon"` is Spanish — "personalizada" slipped in from the `es` row above (the plural key `"Custom Icons"` has the correct "Icônes personnalisées"). `LocalizationTests` won't catch this since the entry exists and is marked `translated`.

```suggestion
            "value" : "Icône personnalisée"
```

After fixing, run `swift scripts/normalize-xcstrings.swift` so the committed bytes match Xcode's canonical serialization.

miquno: Already fixed, and I owe you an explanation for why you were looking at it: this comment is anchored to `acafbe2`, which the branch moved off in a force-push on Aug 4 19:49 — your review landed Aug 5 06:48, so GitHub attached it to a commit that is no longer in the PR (it shows as outdated on the Files tab).

The current head `702f486` has `"Icône personnalisée"`; the only `personali*` values left in the catalog are the three Spanish ones. So there is nothing to change here.

Sorry for the wasted round. I have stopped force-pushing open PRs for exactly this reason — #79 now answers your notes there with two follow-up commits on top of `40356b2`, so your inline anchors stay valid.

PR #79, opened 2026-08-04
feat(entries): download a website's icon into the database

miquno:
> **Stacked on #78.** This branch sits on top of `feat/entry-icon-picker`, so until that merges the diff below shows both commits. Review and merge #78 first; this diff narrows to its own commit automatically once it lands. The commit to read here is `feat(entries): download a website's icon into the database`.

# Summary

Second and final slice of the entry-icon work. #78 could offer the custom icons a database already carried but never add one, so a favicon stayed a cached image on one device — while KeePass's own icon dialog has been downloading them *into the file* for years. This adds the download, storing the result in `Meta/CustomIcons` so the icon travels with the database rather than with the phone, and shows up in every other KeePass client too.

**That section is the whole difficulty.** The parser keeps `Meta/CustomIcons` as one preserved fragment and the writer replays it verbatim, which is exactly why a foreign database's icons survive a KeeForge save untouched. Rebuilding the element from the decoded dictionary would throw that away: `<Icon>` child order, KDBX 4.1's optional `<Name>` and `<LastModificationTime>`, and each writer's own base64 wrapping are all things this app does not model and must not normalize. So `CustomIconXML` splices the new element in ahead of the closing tag and leaves every byte before it alone. Base64 cannot contain `<` or `>`, which is what makes searching the fragment's own text for that tag unambiguous.

Note what this does **not** touch: `KDBXWriter` and `KDBXXMLSerializer` are unchanged. The edit happens in the draft layer, and the writer keeps replaying the fragment exactly as it always has — so a database nobody adds an icon to is byte-for-byte unaffected by this PR.

The written `<Icon>` carries only `<UUID>` and `<Data>`. `<Name>` and `<LastModificationTime>` are 4.1 additions, and writing one would oblige `KDBXWriter.requiredMinorVersion(for:)` to raise the file's minor version — an icon download has no business upgrading the format.

# Changes

- `KeeForge/Models/CustomIconXML.swift` — new: builds the `<Icon>` element and splices it into the preserved fragment, opening the empty `<CustomIcons/>` form by replacing only its `/>` so the start tag survives, and creating the element when the database has none.
- `KeeForge/Models/DatabaseDraft.swift` — `addEntryCustomIcon` stores the image and points the entry at it in one edit (the halves are not independently useful), updating the decoded dictionary and the preserved XML together. An identical image already in the database is reused, so the same favicon on three entries leaves one icon in the file. Shares its entry-side half with `setEntryIcon` so the two cannot drift on where `<CustomIconUUID>` lands.
- `KeeForge/Services/AppSupport/FaviconIconEncoder.swift` — new: redraws at or below 128px (never *up*) and re-encodes as PNG, with a byte cap as a fail-closed backstop. A favicon in a list is fetched and discarded; one in the vault is copied on every save and synced everywhere, so the host does not get to decide what that costs.
- `KeeForge/Services/AppSupport/FaviconService.swift` — `fetchFavicon(for:cachingToDisk:)`. The download does **not** add to the favicon disk cache: that cache is a plaintext per-domain fingerprint of the vault, and the image is about to live somewhere encrypted, so a second copy keyed by domain would widen the fingerprint for nothing. An already-cached favicon is reused, which avoids the request entirely.
- `KeeForge/ViewModels/DatabaseViewModel.swift` — `downloadFavicon(forEntryID:)` and `canDownloadFavicon(forEntryID:)`, with two distinct failures so the user is told which one happened. Runs whatever `showWebsiteIcons` says: that setting governs fetching KeeForge does on its own, and this is the user asking.
- `KeeForge/Views/EntryIconPickerView.swift` — "Download Website Icon" above both grids, disabled when the address names no public host, dismissing on success.
- CHANGELOG, three folder READMEs, and the new strings in `en`, `de`, `fr`, `es`.

# Testing

The claim that matters is not "the splice produces a string" but "the file is still a KDBX and the icon is in it", so the proof is a full write and re-parse:

- **New `add-entry-custom-icon` compatibility scenario** — applies the edit, writes a real KDBX through `KDBXWriter`, parses it back, and asserts the icon arrived byte-for-byte, that every icon the source already held is unchanged, that the count grew by exactly one, that the rest of Meta is untouched, and that the entry carries the `<CustomIconUUID>` reference after the round trip. Registered in both external-expectation allowlists and the artifact count (28 → 29). Like group tags, the icon half stays in-process — `keepassxc-cli` has no verb that prints custom icons — but the artifact still proves the file opens.

```
xcodebuild test -project KeeForge.xcodeproj -scheme KeeForge \
  -destination 'platform=iOS Simulator,name=iPhone 17 Pro' \
  -only-testing:KeeForgeTests/CustomIconXMLTests \
  -only-testing:KeeForgeTests/DatabaseDraftTests \
  -only-testing:KeeForgeTests/DatabaseViewModelTests \
  -only-testing:KeeForgeTests/KDBXRoundTripTests \
  -only-testing:KeeForgeTests/KDBXCompatibilityTests \
  -only-testing:KeeForgeTests/LocalizationTests
```

- 288 tests, 0 failures. `CustomIconXMLTests` pins the splice against a deliberately foreign fragment (other child order, a 4.1 element, whitespace) and the empty-element and no-element cases; `DatabaseDraftTests` covers both Meta representations being written, the entry reference landing after `<IconID>`, dedupe, and a failed edit leaving no orphaned icon.
- `DatabaseViewModelTests` covers the offer/refuse agreement — no URL, `localhost`, a private IP and a `.local` name are all refused before any network call, so the button is never enabled where the action could only fail.
- `xcodebuild build -scheme KeeForgeMac -destination 'platform=macOS' CODE_SIGNING_ALLOWED=NO` — succeeds.

No UI test drives the download itself: it would make a real network request on every CI run. The button's enablement rule is covered at the view-model layer instead.

# Checklist

- [x] Kept AutoFill extension target membership/imports in sync for shared code — `CustomIconXML` lands in `KeeForge/Models`, which both allow-lists take wholesale, and is Foundation-only; `FaviconIconEncoder` is app-target only, since the extension never writes icons. Allow-lists verified byte-identical.
- [x] Updated `KDBXCompatibilityTests` if parser/writer/save behavior changed — new `add-entry-custom-icon` scenario, both expectation allowlists, artifact count
- [x] Preserved accessibility identifiers or updated the affected UI tests — one identifier added (`entry-icon-picker.download-favicon`); existing identifiers unchanged


KeeForge/Services/AppSupport/FaviconIconEncoder.swift
```diff
@@ -0,0 +1,101 @@
...
+    /// crossing it means the encode produced something unexpected, and an
+    /// unexpected blob is not something to write into a vault.
+    static let maximumByteCount = 64 * 1024
+
+    /// PNG bytes for `image`, or nil when it cannot be encoded within the caps.
+    ///
+    /// PNG because it is the format KeePass clients write into `Meta/CustomIcons`
+    /// and the one every reader is certain to decode, and because favicons are
+    /// flat graphics with transparency, which is exactly what it is good at.
+    static func iconData(from image: PlatformImage) -> Data? {
```
crazytan, line 59: `FaviconIconEncoder` is the one new component in this PR with no unit tests, and it's the piece guarding what actually gets written into the vault. All of its contracts are testable in-process with synthetic `CGImage`s — no network, no fixtures:

- the downscale threshold (a 129px image comes back ≤128, a 128px image comes back untouched — the doc comment promises the no-resample path),
- the never-upscale rule (a 16px favicon stays 16px),
- aspect-ratio preservation on non-square input,
- the 64 KB fail-closed cap returning `nil`,
- the exotic-format redraw (indexed / grayscale / 16-bit-per-channel sources — the fixed-BGRA-context comment says these are exactly the cases that motivated it, so they deserve a pin).

Given the PR's own framing that vault bytes are the thing being protected here, I'd close this gap before merge.

KeeForge/Views/EntryIconPickerView.swift
```diff
@@ -0,0 +1,229 @@
...
+            .frame(maxWidth: .infinity)
+        }
+        .buttonStyle(.bordered)
+        .disabled(canDownloadFavicon == false || isDownloading)
+        .accessibilityIdentifier("entry-icon-picker.download-favicon")
+    }
+
+    private func downloadFavicon() {
+        isDownloading = true
+        Task {
```
crazytan: This unstructured `Task` outlives the sheet: if the user pulls the sheet down mid-download, the fetch, the `addEntryCustomIcon` edit, and the save all still happen invisibly behind a dismissed sheet.

That may be acceptable — the user did tap the button — but right now it's an accident of the task's lifetime rather than a decision. Tying the work to the view (e.g. a `.task(id:)`-based pattern, or cancelling the stored task in `onDisappear`) would make dismissal mean "never mind". Either way, worth deciding deliberately; if fire-and-forget is the intended semantic, a line in the doc comment saying so would do.

crazytan (commented):
Two inline notes from review — one test-coverage ask, one behavioral question on the download task's lifetime. Full review discussion in the session; everything else looked good.

KeeForge/Services/AppSupport/FaviconIconEncoder.swift
miquno: Agreed, and done in `25a5104` — `KeeForgeTests/FaviconIconEncoderTests.swift`, 12 tests, all driven by synthetic `CGImage`s with no network and no fixtures. Your five, plus two:

- downscale threshold (129 → 128) and the no-resample path at 128,
- never-upscale (16 stays 16),
- aspect ratio wide, tall, and non-integer-scaling (512×256 → 128×64, 200×400 → 64×128, 300×100 → 128×43, which pins the `.rounded()`),
- PNG output, via the file signature,
- the 64 KB cap failing closed,
- the fixed-BGRA redraw for grayscale, indexed, and 16-bit-per-channel sources.

One finding worth recording, because it decided how the cap test is written: through the downscale path the cap is effectively unreachable. An 8-bit noise image at 128×128 encodes to ~54 KB, and a 256×256 noise image downscaled to 128 lands at ~64.1 KB — under 65536, but with a margin too thin to pin a test on. The one shape that actually reaches it is a **16-bit-per-channel source at exactly the maximum edge**: `downscaled()` returns it untouched, so the 16 bpc survives into the encoder and the PNG comes out around 104 KB. That is the input the cap test uses, and the remaining 8-bit worst case is covered by a separate "whatever comes back is within the cap" assertion.

I measured all the expected numbers against the real encoder before writing the assertions rather than deriving them on paper.

KeeForge/Views/EntryIconPickerView.swift
miquno: You are right that it was an accident rather than a decision. Decided now, in `e7debf2`: **closing the sheet means "never mind"** — but only for the download, never for a write.

- `EntryIconPickerView` holds the task in `@State` and cancels it in `onDisappear`. `CancellationError` is swallowed rather than turned into an alert on a sheet nobody is looking at.
- `DatabaseViewModel.downloadFavicon` calls `try Task.checkCancellation()` after the fetch and before anything reaches the draft. This is needed rather than decorative: `FaviconService.fetchFavicon` does not throw, it returns `nil` on cancellation, so without the check a cancelled download surfaced as a misleading `noIconAvailable`.
- `EntryDetailView` runs the save in `await Task { … }.value`, deliberately outside the picker's task, since an unstructured task does not inherit cancellation. Past the check the icon is already in the draft, and a cloud save *is* cancellable through URLSession — tearing it off half-way would leave the icon applied but unwritten.

So the point of no return is explicit and the sheet sits on the safe side of it.

One gap I want to name instead of hiding: **the cancel path itself has no unit test.** Reaching it means running `downloadFavicon` into `FaviconService.fetchFavicon`, which is a real network call with no injection seam, and adding a seam only for the test is the kind of speculative indirection CLAUDE.md rules out. If you would rather have the coverage, the honest way is a fetch seam on `FaviconService` as its own refactor — say the word and I will do it separately.

crazytan (commented):
Can you rebase to main again?

crazytan (commented):
Review findings on the favicon download flow — the first two are the ones I'd consider blocking; the rest are small in-scope fixes.

KeeForge/ViewModels/DatabaseViewModel.swift
```diff
@@ -986,6 +986,73 @@ final class DatabaseViewModel {
...
+        // it no longer wants.
+        try Task.checkCancellation()
+        guard let image, let imageData = FaviconIconEncoder.iconData(from: image) else {
+            throw FaviconDownloadFailure.noIconAvailable
+        }
+
+        // Downloading the icon an entry already displays is a no-op for the same
+        // reason re-picking the current icon is in `setEntryIcon`: the edit would
+        // cost a history version, a re-encrypt and a save for no visible change.
+        if let current = entry.customIconUUID, unlockedMeta?.customIcons[current] == imageData {
```
crazytan: **Stale `unlockedMeta` after a failed save (blocking).** Every custom-icon read path — `customIconData(for:)`, the picker's `customIcons` grid, and this no-op guard — reads `unlockedMeta`, which is only refreshed on a *successful* save. Offline/conflicted save: the edit lands in the draft, `saveHandlingError()` swallows the failure, and the entry now carries a `customIconUUID` that `unlockedMeta` can't resolve — the entry renders its fallback icon, the picker omits the new image, and the download looks like it did nothing. On retry this guard misses (`unlockedMeta` lacks the UUID), so `.addEntryCustomIcon` re-applies; the draft dedupes the image bytes, but `entryDisplaying` still pushes a fresh `<History>` clone and bumps `lastModificationTime` — N retries write N junk history versions.

Suggested fix: a draft-aware meta accessor (`draft?.meta ?? unlockedMeta`, mirroring the existing `currentRootGroup`) used by all four read sites, plus a test for the failed-save path.

KeeForge/ViewModels/DatabaseViewModel.swift
```diff
@@ -986,6 +986,73 @@ final class DatabaseViewModel {
...
+
+        var image = FaviconService.cachedImage(for: domain)
+        if image == nil {
+            image = await FaviconService.fetchFavicon(for: domain, cachingToDisk: false)
+        }
+        // Point of no return: everything below lands in the draft, so a caller
+        // that gave up during the fetch is told so rather than handed an edit
+        // it no longer wants.
+        try Task.checkCancellation()
+        guard let image, let imageData = FaviconIconEncoder.iconData(from: image) else {
```
crazytan: **Untrusted image decode + PNG encode on the main actor (blocking).** `DatabaseViewModel` is `@MainActor` and `FaviconIconEncoder.iconData` is a synchronous nonisolated call, so the downloaded image's first full decode happens inside `context.draw()` on the main thread — and nothing upstream bounds the *source* pixel dimensions (`maximumEdge` only sizes the destination context). A multi-megapixel favicon freezes the UI for the duration, or spikes memory. `FaviconService.cachedImage` (line 1024) also does synchronous disk I/O + decode here. This runs against the repo rule to keep parsing of untrusted input off the main thread.

`FaviconIconEncoder` is already nonisolated, so hopping off the actor for the cache read + encode (e.g. a nonisolated async helper awaited from here) is a contained fix.

KeeForge/ViewModels/DatabaseViewModel.swift
```diff
@@ -986,6 +986,73 @@ final class DatabaseViewModel {
...
+    /// the entry at it.
+    ///
+    /// A cached favicon is used when one is already on disk, and a fetch does
+    /// *not* add to that cache: the image is about to live in the vault, so a
+    /// second plaintext copy keyed by domain would widen the cache's fingerprint
+    /// for no benefit. Unlike the automatic favicons in lists, this runs
+    /// whatever the "show website icons" setting says — the setting governs
+    /// fetching KeeForge does on its own, and this is the user asking.
+    func downloadFavicon(forEntryID entryID: UUID) async throws {
+        guard isReadOnly == false, let entry = entryIndex[entryID] else { return }
```
crazytan: This guard returns cleanly when the database is read-only or the entry has vanished, and the picker treats any non-throwing return as success — it dismisses and runs the follow-up save with nothing staged, indistinguishable from a stored icon. The window is narrow (a tap racing a read-only flip from another scene, or lock/entry-deletion teardown by one runloop), but throwing a distinct error here keeps failure distinguishable from success for free.

KeeForge/ViewModels/DatabaseViewModel.swift
```diff
@@ -986,6 +986,73 @@ final class DatabaseViewModel {
...
+        guard let domain = FaviconService.extractDomain(from: entry.url) else {
+            throw FaviconDownloadFailure.noPublicDomain
+        }
+
+        var image = FaviconService.cachedImage(for: domain)
```
crazytan: Minor reuse note: this inlines the cached-else-fetch composition that `FaviconService.favicon(for:)` — the service's documented primary API — already is, solely to pass `cachingToDisk: false` to the fetch half. `favicon(for:)` has no settings check, so threading the flag through it (`favicon(for:cachingToDisk:)` defaulting to `true`) is behavior-preserving and keeps the composition in one place.

KeeForge/Services/AppSupport/FaviconIconEncoder.swift
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+    /// Redraws the image at or below `maximumEdge`, preserving aspect ratio.
+    ///
+    /// Returns the original untouched when it already fits, so a favicon that
+    /// arrives at a sane size is never put through a resampling pass that could
+    /// only soften it.
+    private static func downscaled(_ image: CGImage) -> CGImage? {
+        let longestEdge = max(image.width, image.height)
+        guard longestEdge > maximumEdge else { return image }
```
crazytan: This early return skips the normalizing BGRA redraw exactly for the small exotic sources the redraw was written for (per its own comment). A ≤128px 16-bit-per-channel favicon goes straight to `CGImageDestination`, can blow the 64 KiB cap, and surfaces as `noIconAvailable` — "It may not have one, or the connection failed" — for a site that served a usable icon an 8-bit same-size redraw would have fit easily. The test suite pins this refusal as intended, so if it's a deliberate fail-closed tradeoff, fine — but redrawing whenever the source format differs from the 8-bit target (not only when oversized) removes the false refusal, and the pinning test would want updating with it.

KeeForge/Models/DatabaseDraft.swift
```diff
@@ -302,35 +308,89 @@ struct DatabaseDraft: Sendable {
...
+        iconUUID: UUID,
+        imageData: Data
+    ) throws -> (rootGroup: KPGroup, meta: KPMeta) {
+        guard let entryLocation = findEntryLocation(entryID: entryID, in: currentRootGroupStorage) else {
+            throw DraftError.entryNotFound(entryID)
+        }
+
+        var meta = currentMetaStorage
+        let resolvedUUID: UUID
+        if let existing = meta.customIcons.first(where: { $0.value == imageData })?.key {
```
crazytan: `Dictionary.first(where:)` iterates in unspecified per-launch order, and the parser stores foreign icons with no value-level dedup — so a database that already holds identical image bytes under two UUIDs gets a nondeterministic reference: same edit, same input, different file bytes across runs. That contradicts `EntryEdit`'s own "pure function of its input" doc on this case. Damage is low (both UUIDs render identically), but a one-line deterministic tiebreak — e.g. `min(by:)` over the UUID among matching values — restores the documented contract.

KeeForge/Models/CustomIconXML.swift
```diff
@@ -0,0 +1,79 @@
...
+        let closingTag = "</\(elementName)>"
+        if let closing = fragment.range(of: closingTag, options: .backwards) {
+            return fragment.replacingCharacters(in: closing, with: icon + closingTag)
+        }
+        // Opening the empty form replaces only its `/>`, so the start tag —
+        // attributes and all — stays exactly as the other client wrote it.
+        if let selfClosing = fragment.range(of: "/>", options: .backwards) {
+            return fragment.replacingCharacters(in: selfClosing, with: ">" + icon + closingTag)
+        }
+        return "<\(elementName)>\(icon)\(closingTag)"
```
crazytan: Both fallback branches here are unreachable today — `XMLCaptureElement.render()` is the only production path into `OpaqueXMLNodes` and always canonicalizes a preserved fragment to end in a literal `</CustomIcons>` (a source `<CustomIcons/>` is re-rendered as `<CustomIcons></CustomIcons>` before it's preserved, so the `/>` branch and its "ordinary first-icon case" comment don't match what actually reaches this function). The dangerous part is that this last branch is fail-*open*: if a future parser canonicalization change ever made it reachable, it would rebuild the element from scratch and silently drop every existing custom icon in the vault on the next save — the exact loss this type exists to prevent. Suggest failing closed (return the fragment unchanged, or assert) and dropping the dead `/>` branch.

Issue #95, opened 2026-08-16
WebDAV temporarily unavailable - use local copy


xppx99:
### What problem would this solve?

In my setup, WebDAV server is not always available, depending on the location and other firewall rules. 

Currently it tries to reach for the server for some time, and does not let me enter de vault, until there is some kind of timeout. 

Thanks 

### Proposed solution

In this cases I would expect is when KeeForge cannot access the server, anutomatically opens a local copy instead, this being fast and automatic.

crazytan:
Good news: opening from the cached copy when the server can't be reached already existed — the problem was how long it took to give up (30 s per request, and two requests). `main` now gives the server a short window (10 s) to answer the initial "has the file changed?" probe when a local cached copy exists; if it doesn't answer, KeeForge opens the cached copy immediately with the "Using the cached copy offline." notice. Once the server answers, the download itself still uses the normal timeouts. Note that while the server is unreachable, edits can't be saved back yet (the app has no offline upload queue — that's tracked in #67). Will be in the next release.

xppx99:
I think the implemented solution is definitely an improvement. However, in my opinion, it is still not quite perfect.

I believe we could implement something similar to what some alternative applications already do:

- For databases marked as Quick Access: immediately open the cached copy and, in parallel, trigger synchronization with the server. Once synchronization is complete and connectivity is available, the database currently being viewed by the user could be quickly refreshed with the latest version.
- For databases that are not marked as Quick Access: use the mechanism that has now been implemented.

I might even suggest using a slightly shorter timeframe for the second case, but overall, the current implementation is already a good improvement.

PR #126, opened 2026-09-15
feat(cloud): open Quick Launch from the cached copy and sync in the background

### LINKED ISSUE #116: Webdav server - improvements on quick access
### What problem would this solve?

Referring to fix #95 

I think the implemented solution is definitely an improvement. However, in my opinion, it is still not quite perfect.

I believe we could implement something similar to what some alternative applications already do:

- For databases marked as Quick Access: immediately open the cached copy and, in parallel, trigger synchronization with the server. Once synchronization is complete and connectivity is available, the database currently being viewed by the user could be quickly refreshed with the latest version.
- For databases that are not marked as Quick Access: use the mechanism that has now been implemented.

I might even suggest using a slightly shorter timeframe for the second case, but overall, the current implementation is already a good improvement.

### Proposed solution

- For databases marked as Quick Access: immediately open the cached copy and, in parallel, trigger synchronization with the server. Once synchronization is complete and connectivity is available, the database currently being viewed by the user could be quickly refreshed with the latest version.
- For databases that are not marked as Quick Access: use the mechanism that has now been implemented.

### Alternatives considered

_No response_

### Anything else?

_No response_

miquno:
Fixes #116

## What this does

Adds one opt-in setting — **Settings → Cloud Sync → "Open Quick Launch from Cache"**, default **off** — that makes the Quick Launch database open straight from the copy already on this device, and moves the sync behind the unlock instead of in front of it.

#95 bounded the metadata probe so an *unreachable* server costs a fixed deadline rather than a full transport timeout. A *reachable but slow* one still costs that probe plus an entire download before the unlock sheet lets go. This closes that. The deadline itself also drops from 10 s to 5 s — see below.

| Condition | Behavior |
|---|---|
| Setting on + Quick Launch + valid cache | Open the cache immediately, sync in the background |
| Background sync finds nothing new | Cache and session untouched |
| Background sync downloads a newer version | It lands in the cache for the next open; a banner says so; the open database is **not** refreshed |
| Background sync fails or is cancelled | Session stays usable, cache intact, existing error conventions |
| No valid cache / not Quick Launch / setting off / local database | Existing behavior, unchanged |

## Why it is safe to keep the cached copy open and editable

The shared cache is also the **save gate**: `CloudDatabaseSaver.saveOffMain` requires `SaveBaseline.covers(sha512(cache))` before it uploads anything. So when the background sync replaces the cache with a newer remote, the session's next save fails that gate and returns `.conflict` carrying the downloaded bytes — the existing Merge / Save as Copy / Reload flow. Neither side is silently overwritten.

This is load-bearing and worth stating explicitly: the revision gate **cannot** catch this case, because the background sync also rebases the recorded rev, so `remoteHasDiverged` is false. The SHA-512 cache gate runs first and must keep running first. `CloudDatabaseSaverTests.testSaveConflictsWhenABackgroundSyncReplacedTheCacheAfterACachedFirstOpen` pins exactly that, including `uploadCallCount == 0`.

The background pass never touches `rootGroup`, `draft`, or `openTimeSHA512`. Swapping the open database out from under an edit is the data loss this must not cause — which is also why the issue's own suggestion to "quickly refresh the database currently being viewed" is deliberately **not** implemented.

## Provider support

No provider-name checks. Eligibility is capability-shaped: `isCloudBacked` plus a cache whose plaintext KDBX outer header parses. Every `CloudProvider` shares the same cache and the same save gate, so WebDAV, Dropbox and OneDrive are all supported by construction; a provider that does not resolve on the current platform already takes the `.disconnectedCached` cache-first path. Local, iCloud Drive and Files-provider databases are `.local` — they have no cached copy to substitute — and keep their existing behavior.

## The probe deadline now drops to 5 s

This was an open question in the first revision of this PR; it is now decided and implemented, in the direction #116 itself suggested ("a slightly shorter timeframe").

With the new setting on, `CloudSyncCoordinator.openProbeDeadline` is bypassed entirely for the Quick Launch database — there is no probe at open at all. It still governs Quick Launch with the setting off or with no cache, every non-Quick-Launch cloud database, and reload/refresh. Ten seconds is the wait a user sits through **every single time** on a broken connection, with a usable copy of the database already on the device, so it is now **5 s**.

The trade is real and worth stating plainly: a server that is merely slow rather than broken — cold TLS over mobile data, a self-hosted box waking from sleep — and would have answered at, say, seven seconds now loses the race and falls back to the cache. That costs a possibly stale open behind the existing "Using the cached copy offline." banner. It is not a data-safety cost:

- the fallback (`fallbackResolutionIfPossible`) writes `lastSyncError` and nothing else — the recorded rev and content hash are left alone, deliberately, because nothing was learned about the remote;
- so a save from that session still hits `CloudDatabaseSaver.remoteHasDiverged`, which sees the newer remote and raises the ordinary conflict prompt rather than overwriting it;
- and opening or reloading re-probes, so a slow-but-healthy server recovers on the next attempt.

The value is pinned by `CloudSyncCoordinatorTests.testOpenProbeDeadlineIsFiveSeconds`: it is a product decision rather than an implementation detail, and a later change to it should be deliberate and visible in review. Making it user-configurable remains possible but is not in this PR — one more timeout setting is a worse default than a number that is simply short enough.

## Open question for review

**Should cached-first reach more than one database?** `DatabaseListViewModel.toggleQuickLaunch` clears the flag on every other database, so at most one database is ever Quick Launch and this setting therefore applies to exactly one. If that is too narrow, the alternatives are to widen the global setting to all cloud databases, or to add a per-database toggle in Database Details beside "Quick Launch" and "Read Only". Either is a product decision, so it is not in this PR.

## Tests

Regression tests verified to fail before the change and pass after:

- `DatabaseViewModelTests.testCachedFirstOpenUnlocksWithoutWaitingForTheCloudSync`
- `DatabaseViewModelTests.testBackgroundSyncDownloadWarnsWithoutTouchingTheOpenDatabase`

`CloudSyncCoordinatorTests.testOpenProbeDeadlineIsFiveSeconds` pins the shortened deadline. The existing probe-deadline tests from #95 inject their own deadline (0.1–0.2 s) and are unaffected by the new value.

Also added: eligibility matrix (setting off, not Quick Launch, local source, missing/empty/truncated/non-KDBX cache), recorded sync state left untouched, fall-through to the blocking sync, background-sync failure keeps the session usable, repeated opens coalescing onto one pass, lock cancelling the pass and dropping its result, the save-conflict race above, an unchanged-remote save, and the settings default/round-trip/App-Group-isolation.

All fixtures are synthetic or the bundled test databases. No live server, no real credentials.


miquno:
`macos-unit-tests` is red on this PR, and it is the known Xcode 26.3 `actool` crash rather than anything in this branch. Details, since I can't re-run the job myself (no admin rights on this repo).

**What the job does**

The build dies before a single test runs:

```
*** Terminating app due to uncaught exception 'IBPlatformToolFailureException',
    reason: 'The tool closed the connection (AssetCatalogAgent-AssetRuntime)
Command CompileAssetCatalogVariant failed with a nonzero exit code
** TEST FAILED **
error: KeeForgeMacTests: canonical result has no tests
```

That is `ibtoold` crashing while compiling `KeeForge.icon` for the macOS platform — exactly the failure `macos-rc-tests.yml` already documents:

> `macos-15` tops out at Xcode 26.3, whose actool crashes compiling the KeeForge.icon Icon Composer bundle for the macOS platform (it compiles the same .icon for iOS fine).

`macos-rc-tests.yml` dodges it with `runs-on: macos-26` plus an explicit "Expected 26.4+" guard. `pr-tests.yml`'s `macos-unit-tests` job still has `runs-on: macos-15`, where `ls -d /Applications/Xcode_*.app | sort -V | tail -1` selects `/Applications/Xcode_26.3.0.app`. So this PR's macOS job is running on precisely the toolchain the RC workflow was moved off.

**Why it isn't this branch**

- The diff touches no `.icon`, no `.xcassets`, no image, no `project.yml`, no workflow — 13 files, all Swift, docs, `CHANGELOG.md` and `Localizable.xcstrings`.
- `main` has changed nothing icon-related since the last green `macos-unit-tests` run either.
- `xcodebuild build -project KeeForge.xcodeproj -scheme KeeForgeMac -destination 'platform=macOS' CODE_SIGNING_ALLOWED=NO` succeeds locally on Xcode 26.6, asset catalog included.
- `unit-tests` (iOS) is green on the same commit, and it compiles the same catalog for iOS.

**It is intermittent, not deterministic**

Worth knowing if you decide whether to touch the workflow: #125's `macos-unit-tests` passed on the *same* `macos-15` runner with the *same* Xcode 26.3.0 on 2026-09-14 — the catalog compiled and 1719 tests ran green. Here it has now crashed twice in a row, both times at the same build step after ~3 minutes (runs 34947517821 and 35007791132). The exception is a helper process dropping its connection rather than a content error, which fits a flake that Xcode 26.3 hits often but not always.

**Suggested fix, deliberately not in this PR**

Give `pr-tests.yml`'s `macos-unit-tests` job the same treatment `macos-rc-tests.yml` already has: `runs-on: macos-26` and the Xcode 26.4+ guard, so a PR's macOS gate fails loudly on a bad toolchain instead of crashing inside `actool`. Happy to open that as its own small PR if you want it — it seemed wrong to fold a CI change into a feature PR.

A re-run of just this job should be enough to get the PR green in the meantime.

Issue #70, opened 2026-08-02
OTP AutoFill loses subdomain specificity in system suggestions


ftorga:
## What happened?

OTP AutoFill loses the requested subdomain and presents suggestions for the registered domain instead.

For example, when an OTP field belongs to `vt.example.com`, iOS correctly identifies the current context as `vt.example.com`, but the AutoFill UI presents suggestions associated with `example.com`.

The full system picker title references `vt.example.com`, while the suggestions are not specific to that subdomain. Only a limited number of broader-domain suggestions are shown, along with the system action to add a new password.

This differs from the username/password AutoFill flow, where selecting a credential can resolve to the exact KeeForge entry.

## Steps to reproduce

1. Use a KeeForge entry with a valid TOTP configuration, a non-empty title or username, and a website URL such as `https://vt.example.com`.
2. Enable KeeForge as an AutoFill provider.
3. Unlock the database and allow KeeForge to publish AutoFill identities.
4. Open `https://vt.example.com` in Safari on iOS.
5. Focus the site's OTP/verification-code field.
6. Open the full AutoFill suggestions UI.
7. Observe the service context and the displayed suggestions.

## Actual behavior

- The system identifies the active page as `vt.example.com`.
- KeeForge's OTP identity is displayed under the registered domain `example.com`.
- Suggestions are not clearly prioritized for `vt.example.com`.
- Entries associated with the same registered domain may be mixed together.
- The system shows only a limited number of suggestions and offers the generic add-password action.
- When the iOS/iPadOS build runs on macOS through Apple's compatibility layer, no OTP AutoFill option is offered.

## Expected behavior

OTP AutoFill should preserve service specificity:

- credentials for `vt.example.com` should be prioritized first;
- credentials for sibling subdomains should not outrank an exact match;
- a parent-domain fallback should only be used when explicitly relevant;
- multiple credentials for the same exact host should remain selectable.

## Technical evidence

`CredentialIdentityStoreManager.oneTimeCodeIdentity` currently considers only the first valid URL and converts its host to the registered domain before publishing an `ASOneTimeCodeCredentialIdentity`. Therefore `vt.example.com` becomes `example.com`.

Apple documents that the ordered service identifiers passed to `prepareOneTimeCodeCredentialList(for:)` represent specificity, with lower indices being more specific, and that providers should use them to filter or prioritize credentials. KeeForge's current matcher converts the requested identifiers into a `Set`, so that order is not preserved in the provider-side matching path.

[...]

PR #130, opened 2026-09-16
fix(autofill): open the picker when the iOS AutoFill bar's key icon is tapped

### LINKED ISSUE #129: fix(autofill): key icon in the iOS AutoFill bar fills instead of opening the picker
## Problem

On iPhone, focusing a username or password field brings up the system AutoFill bar above the keyboard. KeeForge's best match for the site appears there as a suggestion, with a key icon to its right.

- Tapping the **suggestion** fills that credential. Correct.
- Tapping the **key icon** is the system's "let me pick a different credential" affordance. Today it *also* fills the suggested credential immediately.

So there is no way to reach KeeForge's search from the keyboard bar. A user who keeps two accounts for the same site, or whose best match is simply not the one they want right now, has to let the wrong credential be filled, clear the fields, and start over from the app.

This is about the native iOS AutoFill/QuickType bar, not about anything KeeForge draws inside the page.

## What iOS controls, and what we control

Controlled entirely by iOS, and not addressable from a credential provider:

- whether the AutoFill bar appears, where it sits, and how it is styled
- which credential is promoted into the bar as *the* suggestion, and the ordering
- the key icon's glyph, position, and the fact that it is drawn at all

Controlled by us: **the two taps arrive as different extension entry points**, so they can be told apart.

| User action | Extension entry point | Carries an identity? |
| --- | --- | --- |
| Taps the visible suggestion | `provideCredentialWithoutUserInteraction(for:)`, then `prepareInterfaceToProvideCredential(for:)` if that returns `.userInteractionRequired` | Yes — `ASPasswordCredentialIdentity` for the one credential the user picked |
| Taps the key icon | `prepareCredentialList(for serviceIdentifiers:)` | No — only the service identifiers |

`prepareCredentialList(for:)` exists precisely to display a list for the user to choose from. It never names a credential, because the user has not chosen one yet. Nothing about the requested behavior needs a workaround or an undocumented API.

## Cause

`CredentialProviderCoordinator.presentPasswordMatchesOrFinish()` is the shared tail for both routes. The by-identity route returns early at the top, so everything below it runs only for an interactive list request. That includes:

```swift
// AutoFillExtension/CredentialProviderCoordinator.swift:1416
if matches.count == 1, strictMatches.count == 1, let entry = strictMatches.first {
    completeRequest(with: entry)
    return
}
```

One strict host match means fill and finish, with no picker. Most sites have exactly one saved credential, which is why the key icon looks like it does nothing but fill.

The block dates back to the original coordinator refactor, before the by-identity and list routes were separated. As a convenience for a request that already names its credential it was reasonable; on `prepareCredentialList(for:)` it turns an explicit request to *choose* into an automatic fill.

Ruled out as causes:

- **Over-broad credential metadata.** Password identities intentionally collapse to the registrable domain (eTLD+1); one-time-code identities keep the exact host (#70). Matching is unchanged by this issue.
- **The silent path.** `provideCredentialWithoutUserInteraction` has its own single-match fill, and that one is correct: iOS forbids UI there, so the alternative is `.userInteractionRequired` and an extra hop. It should stay.
- **The picker itself.** It can already search the whole vault and offers "Show All Credentials" (#75). It is simply never presented.

## Expected behavior

- Tapping the visible suggestion keeps filling that credential directly.
- Tapping the key icon opens KeeForge's picker without filling anything first, with the matched entry listed and search available.
- Fields change only after an explicit selection.

## Acceptance criteria

- [ ] Tapping the visible credential suggestion still fills that entry directly.
- [ ] Tapping the key icon does not fill the suggested credential.
- [ ] The key icon opens KeeForge's credential search/picker.
- [ ] The user can search for and select a different entry, including one that did not match the site.
- [ ] Credentials are filled only after an explicit selection.
- [ ] Cancelling or closing the picker leaves every form field untouched.
- [ ] No credential, password, or secret is logged.
- [ ] Existing AutoFill flows are unaffected: silent Quick AutoFill, by-identity fill, passkeys, one-time codes, save-password, and picker-initiated entry creation.
- [ ] Unlock and biometric requirements are unchanged.
- [ ] Verified on the supported iOS range, including the iOS 18 RC simulator target.

## Scope notes

- `presentOTCMatchesOrFinish()` carries the same single-match auto-complete for the one-time-code list request. Same shape, different request type; worth a separate issue rather than widening this one.
- The macOS provider enters the same coordinator, so Safari's picker follows the same change.

miquno:
# Summary

Fixes #129.

This is about the **native iOS AutoFill bar above the keyboard** — the QuickType row with a KeeForge credential suggestion and a key icon to its right — not about anything KeeForge draws inside the page.

Tapping the key icon filled the suggested credential immediately. The key icon is the system's "pick a different credential" affordance, so that left no route from the keyboard to KeeForge's search: the user had to accept the wrong fill, clear the fields, and start over.

Now the key icon opens the picker. Tapping the suggestion itself still fills it directly, unchanged.

# Cause

iOS routes the two taps to different entry points, so they can be told apart:

| User action | Entry point | Identity |
| --- | --- | --- |
| Taps the suggestion | `provideCredentialWithoutUserInteraction(for:)`, then `prepareInterfaceToProvideCredential(for:)` if that escalates | `ASPasswordCredentialIdentity` — a chosen credential |
| Taps the key icon | `prepareCredentialList(for serviceIdentifiers:)` | none — a request to choose |

Both funnel into `presentPasswordMatchesOrFinish()`. Its by-identity branch returns early, so everything after it runs *only* for an interactive list request — including a single-strict-match shortcut that called `completeRequest` and returned. Most sites have exactly one saved credential, so the key icon almost always hit it.

The shortcut predates the split between the by-identity and list routes (it is in the original coordinator refactor). As a convenience for a request that already named its credential it made sense; on `prepareCredentialList` it answers a request to choose with a credential the user never selected.

Not a limitation of iOS, and no workaround involved: `prepareCredentialList(for:)` exists to display a list to choose from.

# Changes

- Removed the single-match auto-complete from the interactive branch of `presentPasswordMatchesOrFinish()`. Six lines; nothing else in the matching logic moved.
- `AutoFillExtension/AGENTS.md`: corrected "A single result fills without a picker", documented the entry-point split, named all three callers that reach the matching code, and stated which lists still auto-fill a lone match.
- `CHANGELOG.md` under `## Unreleased` → `### Fixes`.

Deliberately unchanged:

- **The by-identity path.** Tapping the suggestion still fills directly — covered by the existing `test_passwordFill_resolvesCurrentFormatIdentifier` / `_resolvesLegacyIdentifier`.
- **The silent path.** `provideCredentialWithoutUserInteraction` keeps its own single-match fill. It cannot present anything, so the alternative is an `.userInteractionRequired` bounce for a case that fills correctly today.
- **`presentOTCMatchesOrFinish()`.** The one-time-code list carries the same shortcut. Same shape, different request type, not what was reported — filed as #132 rather than widening this one.
- **Passkey assertion lists.** `presentPasskeyList` still completes on a single match, so a site whose request arrives as a passkey list can still fill without a picker. The CHANGELOG entry says so rather than claiming the key icon always opens one; #132 tracks the fix.

No new UI strings, so no catalog or localization work. No new files, so `project.pbxproj` is unchanged.

# Review follow-ups (442f7a4)

All three points from the review, none of which changes production behavior — the only source edit is a comment.

1. **The branch comment named the wrong callers.** It said everything past the by-identity branch is a list request. Three callers reach it: `prepareCredentialList`, a suggestion whose entry is gone (falls through) or whose database is stale (`resolveInteractiveRequestDatabase`'s `.stale` case clears the target first), and `presentPasskeyList`'s no-passkey fallback. The comment and the `AGENTS.md` line now name all three and say why the picker is right for each.
2. **The CHANGELOG over-claimed.** It read as though the key icon now always opens the picker. `presentPasskeyList` and `presentOTCMatchesOrFinish` still complete on a lone match, so a request arriving as a passkey or one-time-code list still fills without one. Narrowed to passwords, with the exception stated. The remaining work on those two lists is filed as #132.
3. **The macOS shell test used a legacy identifier.** A bare entry UUID parses as `.legacy`. It now builds `CredentialRecordIdentifier(databaseID:entryID:).encoded`, so it takes the branch a real tapped suggestion takes.

The two dependents in point 1 had only multi-match coverage, which never reached the shortcut this PR drops, so their behavior changed untested. Two cases now pin the single-match shape each of them used to fill outright.

# Testing

- `xcodebuild test -scheme KeeForge -destination 'platform=iOS Simulator,name=iPhone 17 Pro' -only-testing:KeeForgeTests` — **1756 passed, 0 failures**.
- macOS, the two classes this touches: `-only-testing:KeeForgeMacTests/CredentialProviderCoordinatorTests -only-testing:KeeForgeMacTests/CredentialProviderShellMacTests` — **80 passed, 0 failures**.
- Xcode 27.0 (27A266a), iPhone 17 Pro simulator on iOS 26.5, macOS host.

The full `KeeForgeMacTests` suite passed locally on the first push (1741 passed, 22 skipped). It no longer completes on this machine: `AppGroupIsolationTests.testClearAllEmptiesTheRedirectedContainerAndLeavesTheRealOneIntact` hangs indefinitely, and it does so when run entirely on its own, on a tree where nothing this branch touches is involved. That is a local environment problem, not a change here; `macos-unit-tests` runs the whole suite on a clean runner.

Seven existing tests failed and none were weakened:

- Five (`test_cleanup_runsOnSuccessfulCompletion`, four `test_copyTOTPOnFill_*`) only used the shortcut as a quick way to reach `completeRequest`. They now select the entry from the picker through a small helper, so each test's actual subject — clipboard behavior, and cleanup running before the credential is handed to the shell — is unchanged.
- `CredentialProviderShellMacTests.test_shellCancel_isNoOpAfterCompletion` did the same to reach a completed request before asserting that a later window close does not re-cancel it. It now names its credential through `targetRecordIdentifier`, the way a tapped suggestion does.
- `test_singleStrictMatch_autoCompletesWithoutOfferingTheOtherEnabledDatabase` pinned the old behavior on purpose ("changing it … should fail here first"). It did. Rewritten as `test_singleStrictMatch_presentsPickerOfferingTheOtherEnabledDatabase`, which also asserts the picker now carries the database switcher — so the second enabled database's entry is reachable, which it was not before.

Six new cases under `// MARK: - Interactive list requests never fill on their own (issue #129)`:

- lone host match is presented, not filled, with the whole vault still in the search corpus
- selecting an entry that never matched the site fills that entry
- cancelling the picker completes no credential at all
- no match: the picker opens on the full corpus with the domain pre-filled, filling nothing
- a stale suggestion with one entry left for the site presents that entry instead of filling it
- the passkey list's no-passkey fallback presents a lone password match instead of filling it

The last two cover the other callers that reach this code. Both routes already had tests, but only in the two-match shape, which never reached the dropped shortcut. With the shortcut restored, each fails on `XCTAssertNil(presenter.completedCredential)`.

# iOS coverage note

The change is unconditional coordinator logic with no `@available` branch, so it behaves identically across the supported range. Only the iOS 26.5 runtime is installed on this machine, so the iOS 18 path is covered by `ios18-rc-tests.yml` rather than locally.

# Checklist

- [x] Ran `xcodegen generate` after adding/removing/moving source files — no files added or removed; project file unchanged
- [x] Updated the nearest folder-local `README.md` if the folder's map changed — `AutoFillExtension/AGENTS.md`
- [x] Localized new UI strings in all shipped locales listed in AGENTS.md and ran `LocalizationTests` — no new strings
- [x] Kept AutoFill extension target membership/imports in sync for shared code — no membership change
- [x] Updated `KDBXCompatibilityTests` if parser/writer/save behavior changed — n/a
- [x] Preserved accessibility identifiers or updated the affected UI tests — none touched


miquno:
Force-pushed one fixup: the first push failed `macos-unit-tests` on `CredentialProviderShellMacTests.test_shellCancel_isNoOpAfterCompletion`, and it was a genuine miss rather than a flake.

That test is inside `#if os(macOS)`, so it only exists in the Mac target, and my local macOS run had been the `CredentialProviderCoordinatorTests` slice alone — which never compiled it. The iOS suite could not catch it either, since the file compiles out there entirely.

The cause is the same pattern as the other updated tests: it used the single-match shortcut only as a way to *reach* a completed request, then asserted that a later window close does not re-cancel it. Its subject is the no-op-after-completion rule, not how the completion was produced. It now names its credential through `targetRecordIdentifier`, which is what a tapped suggestion does and what this change deliberately leaves filling directly. No assertion was relaxed.

Re-verified on the full suites rather than a slice:

- `-only-testing:KeeForgeMacTests` — 1741 passed, 22 skipped, 0 failures
- `-only-testing:KeeForgeTests` — 1754 passed, 0 failures

AutoFillExtension/CredentialProviderCoordinator.swift
```diff
@@ -1411,12 +1411,9 @@ final class CredentialProviderCoordinator {
...
         AutoFillDiagnostics.log("passwordMatches all=\(allPasswordEntries.count) strict=\(strictMatches.count) matches=\(matches.count) possible=\(possibleMatches.count)")
 
-        // Auto-complete without a picker only when the single candidate matched
-        // on host, not on a weaker URL/title substring signal.
-        if matches.count == 1, strictMatches.count == 1, let entry = strictMatches.first {
-            completeRequest(with: entry)
-            return
-        }
+        // Everything below the identity branch is an interactive list request:
+        // the user asked to choose, so even a lone strict match is presented
```
crazytan: This comment (and the matching line in `AutoFillExtension/AGENTS.md`) says only interactive list requests get past the identity branch. Two other callers get here too:

- A by-identity request whose entry is gone. `removeStaleIdentityIfEntryMissing` falls through, and `resolveInteractiveRequestDatabase` sets `targetRecordIdentifier = nil` in its `.stale` case.
- `presentPasskeyList`'s no-passkey fallback, which calls `presentPasswordMatchesOrFinish()`.

Both used to fill a lone strict match and now show the picker. That's probably the better behavior for a stale suggestion, but please describe it accurately so the next reader knows who depends on this branch. The last sentence also describes other paths instead of this code; one short line would do.

crazytan: The same bug is still there on the passkey list route. The extension declares `ProvidesPasskeys`, so on a site that makes a conditional-mediation WebAuthn request (github.com, for example), a key-icon tap can arrive as `prepareCredentialList(for:requestParameters:)`. It then goes to `presentPasskeyMatchesOrFinish` → `presentPasskeyList`, which still calls `complete(entry)` when `matches.count == 1`. A user with a single passkey for that site still gets an immediate fill with no way to choose.

The PR description leaves this out of scope, which is fine, but the CHANGELOG entry says the key icon now always opens the picker. Could you either narrow that wording or open a follow-up issue for the passkey list and the one-time-code list?

KeeForgeTests/CredentialProviderShellMacTests.swift
```diff
@@ -52,9 +52,13 @@ final class CredentialProviderShellMacTests: XCTestCase {
         )
         shell.coordinator.serviceIdentifiers = [serviceIdentifier()]
         seedUnlockedVaultState(shell.coordinator, entries: [entry], sessionKey: sessionKey)
+        // A request that names its credential — the tapped suggestion — is the
+        // one that still fills without a picker; an interactive list request
+        // would present one and complete nothing (#129).
+        shell.coordinator.targetRecordIdentifier = entry.id.uuidString
```
crazytan: Nit: this uses the legacy record-identifier format (a bare entry UUID). It works because `CredentialRecordIdentifier.parse` still accepts `.legacy`, but the identities the app publishes today use the `v2:<databaseID>:<entryID>` format. Building it with `CredentialRecordIdentifier(databaseID:entryID:).encoded` would test the path a real tapped suggestion takes.

AutoFillExtension/CredentialProviderCoordinator.swift
miquno: Right on both counts, and the second one was a behavior change I shipped without noticing. Fixed in 442f7a4.

The comment now names all three callers — `prepareCredentialList`, a suggestion whose entry is gone (the fall-through) or whose database is stale (`resolveInteractiveRequestDatabase`'s `.stale` case, which clears the target before the branch is reached), and `presentPasskeyList`'s no-passkey fallback — and says what they have in common: none of them names a credential the user chose, which is why the picker is right for each. The `AGENTS.md` line says the same. Last sentence trimmed to the one thing it should say.

Worth flagging that both dependents had tests only in the two-match shape, so neither ever reached the shortcut and their change in behavior was unpinned. Two cases now cover the single-match shape each used to fill outright, and both fail on `XCTAssertNil(presenter.completedCredential)` with the shortcut restored.

miquno: Agreed — the CHANGELOG was writing a cheque the change does not cover. Narrowed in 442f7a4: it now says the key icon opens the picker for a **password**, and states the exception plainly, that a site with exactly one saved passkey or verification code still fills without one. `AGENTS.md` says which lists keep the shortcut rather than implying none do.

On the conditional-mediation route specifically: `presentPasskeyList`'s no-passkey fallback into the password flow was also silently filling a lone match, so that path had the reported bug twice over. The fallback half is fixed and pinned by `test_passkeyListFallback_singlePasswordMatch_presentsPickerInsteadOfFilling`. The half you are pointing at — a site where a passkey *does* match — still completes on `matches.count == 1` and is untouched here.

I would rather not widen this PR to the passkey and OTC lists: both need their own decision about the single-match case, and the passkey one has an assertion ceremony behind it. Happy to open the follow-up issue covering `presentPasskeyList` and `presentOTCMatchesOrFinish` — say the word and I will, or resolve this thread if you would rather file it with your own framing.

KeeForgeTests/CredentialProviderShellMacTests.swift
miquno: Fair, and it defeated the point of the line — the test exists to stand in for a tapped suggestion, so it should take the parse branch a tapped suggestion takes, not the compatibility one. It now builds `CredentialRecordIdentifier(databaseID: UUID(), entryID: entry.id).encoded` (442f7a4). Passes on macOS unchanged.

The database UUID is arbitrary here because `entryMatching` resolves on the entry UUID alone for both formats, and this test never goes through database resolution — but `.current` is what it should be exercising either way.

AutoFillExtension/CredentialProviderCoordinator.swift
miquno: Opened as #132 — it covers both `presentPasskeyList` and `presentOTCMatchesOrFinish`, with the reasoning for why each needs its own decision rather than a mechanical repeat of this change (the assertion ceremony on one side, the exact-host identity granularity from #70 on the other). Re-title or re-scope it however you like.

So both halves of your suggestion are done: the CHANGELOG wording is narrowed here, and the remaining work is filed rather than left implicit.

Issue #80, opened 2026-08-07
No option to move item in another group


Fischmuetze:
### What problem would this solve?

There seems to be no option to move item / items in another group (or I can't found it - which would be a hint to inaccessibility) which is very strange

### Proposed solution

have a drag and drop option (in combination with an group edit mode) to move one or multiple items to another group

Issue #118, opened 2026-08-27
Delete Entry doesn't work via search results


dl97ag:
### What happened?

When entry-search has occurred, I'm unable to long-tap to Delete the entry, as tapping Delete does nothing in that scenario.

However manually navigating to the entry (within its Group) allows the Delete to occur without issue.

### Steps to reproduce

Search for target entry so it (and potentially other entries) appear.
Long-tap the entry, and tap Delete ; tapping Delete does nothing.

### KeeForge version

1.15.0

### iOS version and device

iOS 18.7.2 iPhone 13

### Where does the bug occur?

Main app (browsing / unlocking)

### Anything else?

Delete entry works after manually navigating to it:

<img width="1170" height="1407" alt="Image" src="https://github.com/user-attachments/assets/02063c83-44f2-41ab-a18c-c4a99b8e7869" />

Delete entry does not work after searching for it (tapping Delete does nothing):

<img width="1170" height="621" alt="Image" src="https://github.com/user-attachments/assets/b35e2d51-cbc5-46c3-b9b7-0467dc24674d" />

PR #131, opened 2026-09-17
test(ui): cover Move to Group from the search results (#124)

### LINKED ISSUE #124: Move Entry isn’t available via search results
### What happened?

When entry-search has occurred, I'm unable to long-tap to Move the entry to different group, as the “Move to Group” option doesn’t appear.

However manually navigating to the entry (within its Group) allows “Move to Group” option to appear without issue.

### Steps to reproduce

Search for target entry so it (and potentially other entries) appear.
Long-tap the entry, and observe that “Move to Group” option doesn’t appear. 

### KeeForge version

1.15.0

### iOS version and device

iOS 18.7.2 on iPhone 13

### Where does the bug occur?

Main app (browsing / unlocking)

_No response_

### Anything else?

_No response_

miquno:
# Summary

Fixes #124.

**Reported:** on v1.15.0 (iPhone 13, iOS 18.7.2), long-pressing an entry in the search results offered no "Move to Group". Reaching the same entry through its group offered it without issue.

**Expected:** a search result's long-press menu offers Move to Group, and picking a destination moves the entry.

**Root cause.** Move to Group (#80, `7fe94b8`) was wired into `GroupListView`'s own entry-row builder alone. `EntryListView` — the shared list behind both search and the tag browser — never got it, so at tag `v1.15.0` its context menu carries only Delete. That is the reported symptom exactly.

**Current state.** The item reached the shared list with the macOS parity pass (`9b3099a`, 2026-08-22) and shipped in v1.16.0 on 2026-09-13, five days after this issue was filed. I verified the reported flow against `main` on a simulator: the item is offered, the picker presents, and the entry moves. **So this PR changes no behavior** — it adds the regression coverage that was missing, and closes the issue.

The coverage is worth having rather than just closing the issue: Move to Group is the one row action still hand-rolled per shell. `GroupListView`, `EntryListView`, and `MacEntryRow` each build the item and their own `canMove` predicate, unlike the copy pair and Duplicate, which are shared components precisely so they cannot drift. A fourth surface would miss it the same way this one did, and nothing would catch it.

# Changes

- `KeeForgeUITests/EntryEditUITests.swift` — new `SearchResultsMoveUITests`: the reported reproduction (search, long press, Move to Group, pick a destination), asserting the entry both landed in the chosen group and left its old one, plus the cancel path leaving it where it was. The negative assertion is anchored on a sibling entry that stays put, so an old group that has merely not rendered yet cannot read as a successful move.
- `searchForEntries` / `dismissSearch` move from `SearchResultsDeleteUITests` to `EntryEditUITestCase` unchanged, since both search classes now need them.
- `KeeForgeUITests/README.md`, `KeeForge/Views/Entry/AGENTS.md` — document the new class and the per-shell hand-rolling that made #124 possible.

No production code is touched.

# Testing

Run on iOS 26.5 / iPhone 17 Pro simulator:

| Check | Result |
| --- | --- |
| `-only-testing:KeeForgeUITests/SearchResultsMoveUITests` | **2/2 passed** |
| same, against `EntryListView` with the Move item removed (the v1.15.0 shape) | **2/2 failed**, as intended |
| `-only-testing:KeeForgeUITests/SearchResultsDeleteUITests` | **4/4 passed** |
| `SearchResultsMoveUITests` re-run after the final edit | **2/2 passed** |

The deliberate-failure run is the proof the tests bind the reported symptom:

```
XCTAssertTrue failed - Context action 'entry-row.move-context' was not visible for row 'Twitter'
```

`SearchResultsDeleteUITests` is included because the two helpers moved out of it into the base class.

Also run on **iOS 18.6 / iPhone SE (3rd generation)** via a manual `ios18-rc-tests.yml` dispatch ([run](https://github.com/miquno/KeeForge/actions/runs/35194619092)), narrowed to the two search-results classes: **6/6 passed, no retries** — zero failed attempts anywhere in the log, so the pass is not a retry artefact. That covers the reported OS generation and the compact layout, including the short-screen behaviour this class was the open question on.

This had to go through CI because Xcode 27 offers no iOS 18 simulator runtime at all: `xcodebuild -downloadPlatform iOS -buildVersion 18.5`, and the same for `18.7`, both answer `iOS 18.x is not available for download`. The newest runtime installable locally is iOS 26.5. The `macos-15` runners ship the 18.x runtimes preinstalled (Xcode 26.3 there), which is the reason that job exists.

`pr-tests.yml` executes `KeeForgeTests` and `KeeForgeMacTests`, not the UI suite. Its `build-for-testing` step does compile `KeeForgeUITests` through the shared scheme, so the required checks confirm this file builds. Executing them is `ios18-rc-tests.yml`'s job — dispatched manually for this branch above, and automatic at the next RC.

# Checklist

- [ ] Updated `CHANGELOG.md` under `## Unreleased` (feature/bug-fix changes) — n/a, test-only, no behavior change. Worth noting separately: the underlying fix went out in v1.16.0 with no changelog entry of its own.
- [ ] Ran `xcodegen generate` after adding/removing/moving source files — n/a, no files added or moved
- [ ] Localized new UI strings in all shipped locales listed in AGENTS.md and ran `LocalizationTests` — n/a, no new strings
- [ ] Kept AutoFill extension target membership/imports in sync for shared code — n/a
- [ ] Updated `KDBXCompatibilityTests` if parser/writer/save behavior changed — n/a


crazytan:
@miquno Reviewed and merging. The diagnosis is right, the tests are real, and the write-up made this fast to verify — thank you. I checked the load-bearing claims independently rather than taking the PR body at face value:

**Verified**

- **Root cause.** At `v1.15.0`, `KeeForge/Views/GroupListView.swift` carried `entry-row.move-context` and `KeeForge/Views/EntryListView.swift` carried nothing. `git log -S` confirms the item reached the shared list only in `9b3099a`, which is an ancestor of the `644ca7d` folder regroup. Timeline and symptom both check out.
- **Identifiers and fixture facts.** `move-picker.group.<uuid>`, `move-picker.cancel`, `search.entry.navlink`, `search.results.count` and `group.navlink` all resolve. `test.kdbx` is root / `Social` / `Work` / `Work/Internal` / `Empty`, so `label CONTAINS[c] "Work"` has exactly one destination match and `Work` is not Twitter's parent, so the row isn't `disabled`. No helper-name collisions from the hoist.
- **Green.** `SearchResultsMoveUITests` 2/2 passed on an iPhone 17 Pro simulator, 78s total.
- **The tests actually bind to the bug.** I deleted the move item from `EntryListView` and re-ran: both failed with `Context action 'entry-row.move-context' was not visible for row 'Twitter'`. Your deliberate-failure run reproduces exactly.
- **The iOS 18 evidence covers this tree.** `gh api compare` against the fork run's `77426f2` shows zero file differences from the PR head, so that 6/6 is not stale.
- The three `canMove` predicates are currently semantically identical — Mac's `sendToRecycleBin` is just `isEntryInRecycleBin == false` — so nothing has drifted yet.

**The one substantive point, left open**

The PR diagnoses that Move to Group is the last row action hand-rolled per shell, then pins one of three shells with a UI test rather than extracting the shared component. `EntryRowCopyActions.swift` and `EntryRowDuplicateAction.swift` exist precisely so those items cannot drift; an `EntryRowMoveAction` is about fifteen lines, removes the whole failure class including the group-row variant, and costs nothing at test time.

So the doc line — *"A fourth shell would miss it the same way, so `SearchResultsMoveUITests` asserts…"* — overclaims. A search-results UI test protects the shell it drives, not a fourth one. The shared component would. I'm merging the coverage as-is because it stands on its own, and tracking the extraction separately rather than holding this up.

**Smaller items — I'll land these on `main` myself, no action needed from you**

1. Dropping `testCancellingMoveFromSearchResultsLeavesEntryInPlace`. It spends a full launch and unlock (~40s green, ~150s when failing) to assert that `MoveToGroupPickerView`'s Cancel calls `dismiss()` and nothing else. There is no mutation code on that path and the picker is shared by all four surfaces, so a per-shell test adds nothing — unlike #118's cancel case, which guarded a genuine silent-failure bug. It also only asserted the entry stayed in `Social`, not that it hadn't also landed in `Work`.
2. Hoisting the search helpers one level further, to `KeeForgeUITestCase`, built on `activateSearchField` / `clearSearchField` moved down from `UnlockedDatabaseUITestCase`. Landing them on `EntryEditUITestCase` left a third copy of the same search-field dance in the tree next to the sibling base's near-identical pair. That also fixes a real gap: the new `searchForEntries` dropped the `scrollableContainer()?.swipeDown()` reveal both existing copies have, and as a base-class API it will be called by classes that aren't already at the top of the list.
3. Trimming the `KeeForgeUITests/README.md` entry to the what and the identifiers, and letting `Views/Entry/AGENTS.md` carry the why — right now both retell the same v1.15.0-to-parity-pass history at length. While there, naming the gaps the README entry currently implies away: `group-row.move-context` still has no UI coverage anywhere, and the test proves the in-memory list updated rather than that the KDBX location change persisted.

Your `n/a` on the changelog is the right call — the underlying fix shipped in v1.16.0, so a retroactive `## Unreleased` entry would be wrong, and flagging that omission separately was the correct handling.

PR #133, opened 2026-09-17
fix(autofill): open the picker for passkey and one-time-code list requests

### LINKED ISSUE #132: fix(autofill): passkey and one-time-code lists still fill a single match without a picker
## Problem

#129 fixed the AutoFill bar's key icon for **passwords**: an interactive list request now presents the picker instead of filling a lone match. The same shortcut is still in place on the other two list routes, so the reported behavior survives wherever a site's request arrives as a passkey or one-time-code list.

The extension declares `ProvidesPasskeys` and `ProvidesOneTimeCodes` (`AutoFillExtension/Info.plist`, `InfoMac.plist`), so both routes are reachable in normal use:

- On a site making a conditional-mediation WebAuthn request, a key-icon tap arrives as `prepareCredentialList(for:requestParameters:)` and goes to `presentPasskeyMatchesOrFinish` → `presentPasskeyList`. With exactly one saved passkey for that relying party, it completes immediately.
- A one-time-code field's list request arrives as `prepareOneTimeCodeCredentialList(for:)` and goes to `presentOTCMatchesOrFinish`. With exactly one strict host match, it completes immediately.

In both cases the user asked to choose and was answered with a credential they never selected. Having more than one passkey per site, or wanting a verification code from a different entry, is the same dead end #129 describes: accept the wrong one, then start over.

## Cause

Two surviving copies of the shortcut #129 removed from the password path.

```swift
// AutoFillExtension/CredentialProviderCoordinator.swift:1644 — presentPasskeyList
if matches.count == 1, let entry = matches.first {
    complete(entry)
    return
}
```

```swift
// AutoFillExtension/CredentialProviderCoordinator.swift:2211 — presentOTCMatchesOrFinish
if strictMatches.count == 1, let entry = strictMatches.first {
    completeOTCRequest(with: entry)
    return
}
```

Both predate the split between the by-identity and list routes, the same way the password one did.

## Scope

Each needs its own decision rather than a mechanical repeat of #129:

- **Passkeys** carry an assertion, and the list request is parameters-driven. `ASPasskeyCredentialRequestParameters` declares `init` as `NS_UNAVAILABLE` (iOS 26.5 SDK), so `presentPasskeyList(matches:expiredMatches:onSelect:)` is the testable seam and the thin parameters wrapper above it stays untested. Worth checking how the picker should read when the only choice is a single passkey.
- **One-time codes** publish identities at exact-host granularity rather than eTLD+1 (#70), so a lone strict match there is a narrower signal than it is for passwords. It may be that the right answer differs from the password path.

## Deliberately not in scope

`provideCredentialWithoutUserInteraction`'s own single-match fill (`CredentialProviderCoordinator.swift:604`). That path cannot present anything, so the only alternative is a `.userInteractionRequired` bounce for a case that fills correctly today — the same reasoning #129 used to leave it alone.

## Notes

`AutoFillExtension/AGENTS.md` and the `## Unreleased` CHANGELOG entry currently state this limitation explicitly, so both need updating with the fix.

miquno:
# Summary

Fixes #132.

#130 stopped the AutoFill bar's key icon from filling a lone **password** match. The same shortcut survived on the other two list routes, so the reported behavior was still reachable wherever a site's request arrives as a passkey or one-time-code list. The extension declares `ProvidesPasskeys` and `ProvidesOneTimeCodes`, so both are normal-use paths:

- a conditional-mediation WebAuthn request with exactly one saved passkey for the relying party asserted it immediately;
- a verification-code field with exactly one strict host match generated its code immediately.

Both answered a request to *choose* with a credential the user never selected.

# Cause

Two surviving copies of the shortcut #130 removed from the password path:

- `presentPasskeyList` — `if matches.count == 1 { complete(entry); return }`
- `presentOTCMatchesOrFinish` — `if strictMatches.count == 1 { completeOTCRequest(with: entry); return }`

Both predate the split between the by-identity and list routes, the same way the password one did. Neither is reachable from a request that names a credential: `presentPasskeyList`'s only caller is the parameters-driven `prepareCredentialList(for:requestParameters:)`, and `presentOTCMatchesOrFinish` is reached only from `prepareOneTimeCodeCredentialList` and from `completeOTCRequestFromPending`'s stale-identity fallback. So every request that reaches either one is a request the user still has to answer.

# Changes

Each route got its own decision rather than a copy of #130's, because the pickers they reach differ.

- **Passkeys.** Shortcut dropped; a lone match falls into the existing `presentSearchView(entries: matches, includesDatabaseSwitcher: true)` branch. The corpus stays narrowed to the matches rather than widening to the whole vault the way the password picker does — an assertion is only valid for the requesting relying party, so there is nothing else to offer. What the picker adds is the **database switcher**, which is what makes a passkey for the same site in another database reachable at all; completing outright ended the request before it could be. A one-row picker is an already-shipped shape here — the expired-passkey branch presents exactly that.
- **One-time codes.** Shortcut dropped; a lone strict match falls into the existing `entries: matches, searchEntries: allTOTPEntries` branch, so the match is listed and the rest of the TOTP corpus stays reachable through search and "Show All Credentials". No new branch was needed: the picker's shape for a lone match is the one it already had for several.
- **`CredentialMatcher.orderedStrictMatchedEntries` goes with the shortcut.** It existed to pick the single one-time-code entry that was safe to fill, and `presentOTCMatchesOrFinish` was its only production caller — leaving it would have been dead code. #70's exact-host granularity is unaffected: that lives in `CredentialIdentityStoreManager.oneTimeCodeIdentities`, which publishes one identity per exact host and is untouched here. Its five `CredentialMatcherTests` cases go with it; the broad-matcher behavior two of them also asserted is already pinned by `testStrictDoesNotMatchHostSubstring` and `testStrictIsSubsetOfFuzzy`. That second one is the invariant this change leans on — strict ⊆ broad, so a lone strict match is always inside `matchedEntries` and reaches the narrowed picker rather than the no-match branch.
- `AutoFillExtension/AGENTS.md` — the bullet #130 added said the passkey and one-time-code lists still auto-fill a lone match; it now states the rule holds for every interactive list, names the by-identity routes that still complete directly, and records that `strictMatchedEntries` gating the silent no-UI fill is the one place left where a *match* becomes a credential on its own.
- `CHANGELOG.md` — #130's `## Unreleased` entry carried the same exception ("Passkeys and verification codes are unchanged for now"). Amended rather than duplicated, since both are unreleased and one sentence is the truthful user-facing statement.

Deliberately unchanged:

- **The by-identity routes.** `completeInteractivePasskeyRequest` and `completeOTCRequestFromPending` still complete directly — those name the credential the user tapped.
- **The silent no-UI path.** `provideCredentialWithoutUserInteraction`'s own single-match fill stays, per #132's scope note: it cannot present anything, so the only alternative is a `.userInteractionRequired` bounce for a case that fills correctly today.

No new UI strings, so no catalog or localization work. No new files, so `project.pbxproj` is unchanged.

# Testing

Reproduced first: the seven new/rewritten coordinator cases were run against the unmodified tree and all seven failed, each on the shortcut — the one-time-code ones with a generated code (`"402402"`, `"515671"`) where the picker should have been, the passkey ones with an assertion and no `searchView`.

After the fix:

- `xcodebuild test -scheme KeeForge -destination 'platform=iOS Simulator,id=…iPhone 17 Pro' -only-testing:KeeForgeTests` — **1755 passed, 0 failures, 0 skipped**.
- `xcodebuild test -scheme KeeForgeMac -destination 'platform=macOS' CODE_SIGNING_ALLOWED=NO -only-testing:KeeForgeMacTests/CredentialProviderCoordinatorTests -only-testing:KeeForgeMacTests/CredentialProviderShellMacTests -only-testing:KeeForgeMacTests/CredentialMatcherTests` — **120 passed, 0 failures**.
- Xcode 27.0 (27A266a), iPhone 17 Pro simulator on iOS 26.5, macOS host.

Both totals are net of the five deleted matcher cases (1760 → 1755, 125 → 120 on those three Mac classes).

Three coordinator tests pinned the old behavior directly and are rewritten; no assertion was relaxed:

- `test_passkeyList_singleMatch_completesDirectly` → `…_presentsPickerInsteadOfAsserting`. Same setup, and it still asserts the passkey — after an explicit selection.
- `test_otcList_singleMatch_completesWithCode` → `…_presentsPickerInsteadOfFilling`. Still asserts a valid six-digit code, produced by picking the entry, and additionally that the rest of the TOTP corpus is in the search set.
- `test_otcList_prefersMostSpecificOrderedServiceIdentifier` → `test_otcList_nestedSubdomainIdentifiers_offerBothHostMatchesWithoutFilling`. Its subject was *which entry won an automatic fill*, which no longer exists on this route; it now pins that both host matches under the requested domain are offered and neither is filled.

Four new cases cover what the fix adds:

- cancelling the one-passkey picker asserts nothing and cancels with `.userCanceled`
- cancelling the one-match one-time-code picker completes no code
- the one-passkey picker carries the database switcher listing both enabled databases — the concrete gain over completing outright
- `completeOTCRequestFromPending`'s stale-identity fallback presents its lone remaining TOTP match instead of turning it into a code (that caller previously had only multi-match coverage, which never reached the shortcut)

## Not tested, and why

- **The full `KeeForgeMacTests` suite does not complete on this machine.** `AppGroupIsolationTests.testClearAllEmptiesTheRedirectedContainerAndLeavesTheRealOneIntact` hangs indefinitely, on its own and on an unmodified tree — a local environment problem, not this change. `macos-unit-tests` runs the whole suite on a clean runner. The three classes above are the ones this change can affect.
- **iOS 18.** The change is unconditional coordinator logic with no `@available` branch, so it behaves identically across the supported range. Only the iOS 26.5 runtime is installed here, so the iOS 18 path is covered by `ios18-rc-tests.yml`.
- **No on-device pass.** Reaching these two routes for real needs a site issuing a conditional-mediation WebAuthn request or a one-time-code field, on a device with the extension enabled. Manual check, once available: with exactly one saved passkey for a site, tap the key icon in the AutoFill bar — the KeeForge picker should open listing that passkey, with the database switcher present when a second database is enabled, and the assertion should happen only after tapping the row; cancelling should leave the page untouched. Repeat on a verification-code field with exactly one saved TOTP entry for the host.

# Checklist

- [x] Ran `xcodegen generate` after adding/removing/moving source files — no files added or removed; project file unchanged
- [x] Updated the nearest folder-local `README.md` if the folder's map changed — `AutoFillExtension/AGENTS.md`
- [x] Localized new UI strings in all shipped locales listed in AGENTS.md and ran `LocalizationTests` — no new strings; `LocalizationTests` ran as part of the full iOS suite
- [x] Kept AutoFill extension target membership/imports in sync for shared code — no membership change
- [x] Updated `KDBXCompatibilityTests` if parser/writer/save behavior changed — n/a
- [x] Preserved accessibility identifiers or updated the affected UI tests — none touched


crazytan:
@miquno can you rebase this to main?

PR #135, opened 2026-09-18
refactor(ui): share the entry row's Move to Group item across shells

### LINKED ISSUE #134: Extract a shared EntryRowMoveAction so Move to Group stops being hand-rolled per shell
## Problem

Move to Group is the last entry-row context-menu action that each shell builds by hand. `KeeForge/Views/Group/GroupListView.swift`, `KeeForge/Views/Entry/EntryListView.swift`, and `MacEntryRow` in `KeeForge/Views/Entry/MacEntriesList.swift` each declare their own `Button("Move to Group")`, their own `entry-row.move-context` identifier, and their own `canMove` predicate.

The copy pair and Duplicate are not built this way. `EntryRowCopyActions.swift` and `EntryRowDuplicateAction.swift` exist as shared components precisely so their wording, gating, and identifiers cannot drift per shell.

This is the root cause of #124. Move to Group was added in #80 to `GroupListView`'s row builder alone, so the shared `EntryListView` behind search and the tag browser never offered it, and v1.15.0 shipped with the item missing from search results entirely. It reached the shared list only as a side effect of the macOS parity pass in `9b3099a` and has worked since v1.16.0.

#131 covered the search-results shell with `SearchResultsMoveUITests`, which is worth having, but it holds one shell. A fourth entry-row surface would miss the item the same way and no test would catch it.

## Current state

The three predicates are semantically identical today, so there is no user-visible bug to fix:

- `GroupListView.canMoveEntry` and `EntryListView.canMove`: `isReadOnly == false && isEntryInRecycleBin(entryID:) == false`
- `MacEntryRow.canMove`: `isReadOnly == false && sendToRecycleBin`, where `sendToRecycleBin` is `isEntryInRecycleBin(entryID:) == false`

This is drift risk, not drift.

## Proposed change

Extract an `EntryRowMoveAction` into `KeeForge/Views/Entry/`, beside `EntryRowCopyActions.swift` and `EntryRowDuplicateAction.swift`, and embed it in all three shells.

- Own the item's title, the `entry-row.move-context` identifier, and the single `canMove` predicate.
- Report the choice back through a closure rather than presenting, the way `EntryRowDuplicateAction` hands its editor back, because where the move sheet is hosted differs per shell. `PendingMove` already lives in `PendingRowActions.swift` and the hosts already differ: `EntryListView` and `MacEntriesList` host list-scoped, while the macOS group column raises it to `RegularDatabaseWorkspaceView`.
- Keep the existing identifier so `SearchResultsMoveUITests` and the Mac UI tests keep binding without changes.

Worth deciding in the same change: whether the group-row variant (`group-row.move-context`, `canMoveGroup` in `GroupListView` plus `canMove` in `RegularDatabaseWorkspaceView`) folds into the same component or stays separate. Its predicate genuinely differs, since it adds the deletion-protection screen, so a shared component may not fit as cleanly there.

## Notes

- `KeeForge/Views/Entry/AGENTS.md` already documents this asymmetry and names `EntryRowMoveAction` as the fix that retires it. Update that paragraph when the extraction lands.
- The group-row move has no UI coverage on any surface, which `KeeForgeUITests/README.md` now records as a known gap.
- No user-facing behavior change, so no CHANGELOG entry is expected.

miquno:
# Summary

Fixes #134.

Move to Group was the last entry-row context-menu action each shell built by hand. `GroupListView`, `EntryListView`, and `MacEntryRow` each declared their own `Button("Move to Group")`, their own `entry-row.move-context` identifier, and their own `canMove` predicate. This is how #124 happened. The three predicates were identical, so this was drift risk rather than drift, and **no user-visible behaviour changes**.

# Changes

- **New `KeeForge/Views/Entry/EntryRowMoveAction.swift`**, beside `EntryRowCopyActions` and `EntryRowDuplicateAction`. It owns the item's title, the `entry-row.move-context` identifier, and the single gate `EntryRowMoveAction.isAvailable(entryID:viewModel:)`: the database is writable and the entry is not in the recycle bin. It hands a `PendingMove` back through `onMove` rather than presenting, as the issue proposes. So each shell keeps hosting the sheet where it already did: list-scoped in `GroupListView` / `EntryListView` / `MacEntriesList`, raised to `RegularDatabaseWorkspaceView` from the macOS group column.
- **All three shells embed it.** Their `canMoveEntry` / `canMove` predicates are deleted.
- **`EntryRowDuplicateAction` reuses the same gate** (plus the session key) instead of restating it. Its doc comment already said "same eligibility as Move to Group", and now the code enforces that. This goes slightly beyond the issue. It is a two-line change, easy to drop if you'd rather keep Duplicate independent.
- **Group-row move stays separate. This is the decision the issue left open.** Its gate is `canEditGroup` / `canEdit` plus the deletion-protection screen, and Edit Group and Change Icon share `canEditGroup`. Folding it into an entry-row component would split that predicate away from its siblings.
- **Docs:** replaced the "still hand-rolled per shell" paragraph in `Views/Entry/AGENTS.md` with an entry for the new file, including the group-row decision. Also updated the two sentences in `Views/Group/AGENTS.md` that described the item as hand-built.
- `project.pbxproj`: only the additive lines for the new file. xcodegen's `targets` reordering is reverted.

The identifier is unchanged, so `SearchResultsMoveUITests` binds without changes. For the record, no `KeeForgeMacUITests` case names `entry-row.move-context`. No CHANGELOG entry, because nothing changes for users.

# Testing

New `testEntryRowMove…` cases in `KeeForgeTests/DatabaseViewModelTests.swift` pin each condition of the gate: offered for a writable entry, withheld for a recycled entry, withheld in a read-only database.

| Check | Result |
| --- | --- |
| iOS 26.5 / iPhone 17 Pro: the three new cases | **3/3 passed** |
| same, with the read-only condition removed from the gate | `testEntryRowMoveIsWithheldInAReadOnlyDatabase` **failed**, as intended |
| same, with the recycle-bin condition removed from the gate | `testEntryRowMoveIsWithheldForARecycledEntry` **failed**, as intended |
| `-only-testing:KeeForgeUITests/SearchResultsMoveUITests` | **passed** |
| `-only-testing:KeeForgeUITests/EntryDuplicateUITests` (gate now shared) | **passed** |
| macOS (`CODE_SIGNING_ALLOWED=NO`): `-only-testing:KeeForgeMacTests/DatabaseViewModelTests` | **204/204 passed**, includes the new cases; `KeeForgeMac` builds |

**What is not covered:**
- The unit tests pin the gate, not that each shell renders the component. That the item appears on the iOS group list and on the Mac rows rests on all three embedding the same view. Only the search-results shell has a UI test (`SearchResultsMoveUITests`).
- The group-row move still has no UI coverage on any surface, as `KeeForgeUITests/README.md` already records.
- The full `KeeForgeMacTests` suite was not run locally. `AppGroupIsolationTests.testClearAllEmptiesTheRedirectedContainerAndLeavesTheRealOneIntact` hangs on this machine on an unmodified tree too, so that is left to `macos-unit-tests`.

# Checklist

- [ ] Updated `CHANGELOG.md` under `## Unreleased` (feature/bug-fix changes) — n/a, refactor with no behavior change
- [x] Updated the nearest folder-local `README.md` if the folder's map changed (`Views/Entry/AGENTS.md`, `Views/Group/AGENTS.md`)
- [ ] Localized new UI strings in all shipped locales listed in AGENTS.md and ran `LocalizationTests` — n/a, the "Move to Group" key is unchanged
- [ ] Kept AutoFill extension target membership/imports in sync for shared code — n/a, the file is app-only like its siblings
- [ ] Updated `KDBXCompatibilityTests` if parser/writer/save behavior changed — n/a


crazytan (approved):
Reviewed the shared EntryRowMoveAction refactor. The three entry-row shells preserve the existing availability behavior and accessibility identifier, the group-row action remains appropriately separate, and the focused tests cover the centralized gate. No blocking findings.

Issue #115, opened 2026-08-23
feat: Add (s)FTP support


StanoRiga:
### What problem would this solve?

Right now there are 4 options to connect to a kdbx file. Via the files app, Dropbox, OneDrive and WebDAV. I know that in theory you can connect to a lot more online spaces through files app, but nearly every time KeePass fails to access the file due to very poor implementation by Apple when it comes to online storage integration in files app.

### Proposed solution

A direct connection would solve these issues.
In my case, my kdbx would be reachable via FTP. 
So instead of finding the next app that tries to integrate an FTP storage in files (and will not be reliable), I would love to see the option to directly connect to a FTP.
Thank you for your hard work.

miquno:
FTP support is implemented in #138.

A database can be opened, saved, and created directly on an FTP server, next to WebDAV, with no Files-app provider in between. The client is built on Network.framework, so it adds no dependency. Saves upload to a hidden temporary file and rename it over the database after re-checking the recorded revision, so the remote copy is never written in place. Plain FTP sends the login unencrypted, so the connection has to be turned on explicitly (**Allow Unencrypted FTP**), the same way WebDAV asks before it uses HTTP. Available on iOS and macOS.

SFTP is not part of that PR, which is why this issue stays open. It needs an SSH implementation, and the available Swift options each come with a dependency trade-off that is worth deciding on its own, so I will follow up on that separately.

PR #136, opened 2026-09-18
feat(cloud): open and save databases on FTP servers

miquno:
# Summary

Part of #115. Adds FTP as a storage provider, so a database can be opened, saved, and created directly on an FTP server instead of going through a Files-app integration. SFTP and FTPS are **not** included (see Limitations), which is why this does not close the issue.

# Changes

- **Provider:** `FTPCloudProvider` plugs into the existing `CloudProvider` pipeline (cache, conflict gate, backups, conflict copies, pending AutoFill uploads) the same way WebDAV does. It is stateless and reads its credential from `CloudTokenStore` on each call.
- **Client:** `FTPClient` is built on Network.framework, so there is **no new dependency**.
  - Each operation opens its own logged-in session.
  - Data connections use passive mode: EPSV, falling back to PASV. The data connection always goes to the control host, never the address PASV advertises.
  - Listing and metadata use MLSD/MLST, falling back to LIST, SIZE and MDTM (vsftpd supports neither MLSD nor MLST).
  - Every read and write has a timeout, and task cancellation cancels the connection, so the 10 s open probe works.
  - Command arguments containing CR, LF or NUL are rejected.
- **Saves:** the upload goes to a hidden temporary name, the database is downloaded again and its hash compared with the recorded rev, then the upload is renamed over the database. The database is never written in place or deleted first.
  - A server that cannot rename onto an existing file (IIS): the database is renamed to a `.keeforge-backup` beside it, re-checked, and replaced. Any failure renames it back, over a fresh connection if the old one died.
  - Every failure after `STOR` deletes the temporary upload. Uploads that outlive a dead connection carry their creation time in the name and are swept by a later save in the same folder after 24 hours.
  - `createFile` refuses an existing name before and after the transfer.
- **Existence checks:** a 550 to `MLST` does not count as "missing", because RFC 3659 also allows it for denied access. `SIZE` and then the parent listing must show the name is absent; otherwise the operation fails as permission denied.
- **Rev:** `sha256:<hex>` of the file bytes. `download` and `upload` report the hash of the bytes they transferred, never a re-read.
- **Connect form:** `FTPConnectView`/`FTPConnectViewModel` take an `ftp://host[:port]/path` address, a username and a password. The path is relative to the login directory, RFC 1738 style. Plain FTP sends the login unencrypted, so connecting requires turning on **Allow Unencrypted FTP**, like WebDAV's HTTP opt-in. The browser picks the right form through a small `ManualConnectSheet`.
- **Reconnect:** the Reconnect action explains how to add the server again instead of reporting a configuration error.
- **Platforms:** FTP is available on iOS and macOS. Both targets already have outbound network access.
- **New Database:** the "Save To" picker is now a menu. With five destinations, the segmented control truncated "Local Files" on iPhone.
- **Strings and docs:** new strings are translated into all shipped locales. `NSLocalNetworkUsageDescription` now mentions FTP. Folder docs and CHANGELOG are updated.

# Testing

- New `FTPCloudProviderTests` (50) drive the real protocol code against an in-memory FTP server (`Support/FakeFTPServer.swift`). They cover:
  - connect and persist, rejected login, missing folder, and the opt-in check;
  - MLSD listing, and LIST+PASV listing against the control host;
  - metadata and its SIZE/MDTM fallback;
  - temp-then-rename upload, the post-transfer rev conflict, and the IIS fallback;
  - create-only refusals, before and during the upload;
  - 552/550 mapping, timeout reported as offline, connection refused, CR/LF injection, and the reconnect message;
  - after review: the hash rev (a same-second, same-size write conflicts), existence behind an ambiguous 550, the IIS set-aside and restore (including a dropped connection and a lost reply), temporary-upload cleanup, and the stale-upload sweep. The fake server can now drop the connection before a command or with its reply.
- After the review fixes: FTP, FTP parser/URL/connect-form, `CloudSyncCoordinatorTests` and `CloudProviderRegistryTests` pass on iOS (**116/116**) and macOS (**113 run, 0 failed, 43 Keychain-dependent skipped**; all of those ran on iOS).
- New `FTPListingParserTests`/`FTPURLTests` and `FTPConnectViewModelTests`; `CloudProviderRegistryTests` and `CloudFileBrowserViewModelTests` updated.
- `xcodebuild test -scheme KeeForge -only-testing:KeeForgeTests -only-testing:KeeForgeUITests/DatabaseCreationUITests` on iPhone 17 Pro (iOS 26.5): **1808/1808 passed**.
- `xcodebuild test -scheme KeeForgeMac -destination 'platform=macOS' CODE_SIGNING_ALLOWED=NO` with the FTP, registry, browser, localization, WebDAV, cloud-sync, saver and creation classes: **193 passed, 0 failed, 30 skipped**. The skips are Keychain-dependent tests that skip in an unsigned local host; all of them ran and passed on iOS.
- **Manual checks** against local vsftpd and pure-ftpd servers in Docker (not part of the PR):
  - A temporary test ran the real Network.framework transport end to end on iOS and macOS: create, list, metadata, download, upload with rev, stale-rev conflict, create conflict, wrong password, refused port, and cancellation of an unanswered connect in under 1 s.
  - In the iOS app:
    - add an FTP database, unlock it, and save; the server bytes match the local cache;
    - an external change triggers the Save Conflict dialog, and "Save as Conflict Copy" writes the sibling copy;
    - create a new database on FTP.

# Limitations

- **No SFTP or FTPS.** SFTP needs an SSH library; the only maintained Swift option pulls in a personal fork of swift-nio-ssh, which needs a maintainer decision. Explicit FTPS (AUTH TLS) cannot be done on Network.framework.
- **Extra download per save.** The rev is a content hash, so `getMetadata` and every save download the database once more. Cheap at KDBX sizes, but not free.
- **Last-moment writes.** A write that lands between the hash check and the rename is still lost on servers whose rename overwrites; FTP has no conditional write.
- **Sweep needs dotfiles.** Leftover uploads are only swept where the listing shows hidden files (MLSD does; vsftpd's `LIST` does not).
- **Backup can remain.** If the connection dies while the IIS path has the database set aside and something already stands at the original name, the `.keeforge-backup` is left in place rather than overwrite it.
- **Create-only is not atomic.** FTP has no exclusive create, so a file appearing between the last existence check and the rename can be replaced on servers whose rename overwrites.
- **Permissions after save.** The rename puts a new file in place, so the saved database gets the server's permissions for new uploads (umask), not its previous mode.
- **Untested setups:** no UI test double for FTP, and the Mac UI was not exercised by hand. IIS and PASV-only servers are covered only by the fake server.

# Checklist

- [x] Updated `CHANGELOG.md` under `## Unreleased`
- [x] Ran a `KeeForgeMacTests` slice when shared or Mac behavior changed
- [x] Updated the nearest folder-local docs if the folder's map changed
- [x] Localized new UI strings in all shipped locales and ran `LocalizationTests`
- [x] Kept AutoFill extension target membership/imports in sync for shared code (no extension changes)
- [ ] Updated `KDBXCompatibilityTests` if parser/writer/save behavior changed (not needed: no parser/writer change; the save pipeline is reused unchanged)


KeeForge/Services/Cloud/FTPCloudProvider.swift
```diff
@@ -0,0 +1,392 @@
...
+    private static func replace(
+        _ path: String,
+        with temporaryPath: String,
+        data: Data,
+        in session: FTPSession
+    ) async throws {
+        do {
+            try await session.rename(temporaryPath, to: path)
+        } catch FTPClientError.unexpectedReply(let reply) where reply.code == 550 || reply.code == 553 {
+            try await session.store(path, data: data)
```
crazytan: **P1 — preserve the existing database on rename fallback.** `STOR path` retransmits directly over the live database. On servers that truncate or open the destination before consuming all data, a timeout or disconnect during this second transfer leaves the user's remote copy partial even though the temporary upload already completed. Please use a preservation/rollback sequence (for example, rename the old target aside, promote the completed temporary file, and restore on failure) or fail the save instead of overwriting in place. The current fake publishes bytes only from `finishSending()`, so it cannot catch this failure mode.

KeeForge/Services/Cloud/FTPClient.swift
```diff
@@ -0,0 +1,529 @@
...
+    func stat(_ path: String) async throws -> FTPListEntry? {
+        let mlst = try await command("MLST", path)
+        if mlst.code == 250 {
+            let entries = mlst.lines.dropFirst().compactMap(FTPListingParser.parseMachineEntry)
+            guard let entry = entries.first else {
+                throw FTPClientError.malformedReply(mlst.message)
+            }
+            return entry
+        }
+        if mlst.code == 550 {
```
crazytan: **P1 — an `MLST 550` reply does not prove absence.** [RFC 3659](https://www.rfc-editor.org/rfc/rfc3659.html) allows 550 when access to directory information is denied as well as when the path is missing. Returning `nil` lets `createFile`'s two existence checks pass on upload-only or ACL-restricted servers; if `STOR`/`RNTO` are allowed, the final rename can replace an existing database. Fall back to `SIZE`/listing and fail closed when absence cannot be established.

KeeForge/Services/Cloud/FTPCloudProvider.swift
```diff
@@ -0,0 +1,392 @@
...
+            return .unknown("\(reply.code) \(reply.message)")
+        }
+    }
+
+    // MARK: - Rev
+
+    /// `mtime:<stamp>;size:<bytes>`, or nil when the server reports no
+    /// modification stamp. Stamps are usually whole seconds, so two writes of
+    /// the same size inside one second are indistinguishable.
+    static func rev(for entry: FTPListEntry) -> String? {
```
crazytan: **P1 — this revision can miss real conflicting writes.** On servers with whole-second MDTM precision, two same-sized KDBX saves within one second produce the same rev. The post-transfer check then accepts the stale writer and replaces the first save. KDBX re-encryption can preserve total length, so this is a credible lost-update path. This needs a strong validator (a server hash or remote-byte hash against the baseline), or writable sync should fail closed when one is unavailable.

KeeForge/Services/Cloud/FTPCloudProvider.swift
```diff
@@ -0,0 +1,392 @@
...
+        data: Data,
+        expectedRev: String?,
+        progress: @escaping @Sendable (Double) -> Void
+    ) async throws -> CloudFileMetadata {
+        let (location, credential) = try resolveContext(accountId: accountId)
+        let path = Self.remotePath(base: location.basePath, fileId: fileId)
+        let temporaryPath = Self.sibling(of: path, named: makeTemporaryName(FTPListingParser.lastComponent(of: path)))
+
+        let metadata = try await perform(context: .write, location: location, credential: credential) { session in
+            try await session.store(temporaryPath, data: data)
```
crazytan: **P2 — clean up the temporary upload on every failure after `STOR`.** Only the explicit revision-mismatch branch deletes it. If the subsequent stat, rename, or control reply fails, the hidden UUID-named file remains and no later operation sweeps it. Repeated transient failures can consume the server's quota. Wrap the post-store sequence in a catch with best-effort deletion, and provide recovery or sweeping for failures that lose the control connection.

crazytan (changes requested):
Requesting changes for four data-safety issues in the FTP implementation. The focused iOS and macOS suites pass, but they do not exercise these destructive or fail-closed edge cases.

KeeForge/Services/Cloud/FTPCloudProvider.swift
miquno: Fixed in 11de86b. The in-place `STOR` is gone. When RNTO refuses to replace, the database is renamed to a `.keeforge-backup` beside it, re-checked against `expectedRev`, and the upload is renamed into place. Any failure renames the backup back, over a fresh session if the control connection died. If something already stands at the original name (for example an upload installed by a rename whose reply was lost), the backup is left in place instead of overwriting it. Tests: `testFailedInstallRestoresTheOriginalDatabase`, `testDroppedInstallIsRestoredOverAFreshConnection`, `testLostInstallReplyKeepsTheBackupRatherThanReplacingTheUpload`.

KeeForge/Services/Cloud/FTPClient.swift
miquno: Fixed in 11de86b. `stat` now returns nil only once absence is shown: after a 550 it asks `SIZE`, then lists the parent folder. If neither settles it, it throws `accessDenied`, which maps to `permissionDenied`. A `LIST` that hides dotfiles cannot prove a hidden name absent. `createFile` therefore fails closed on ACL-restricted servers. Tests: `testCreateFileNeverReplacesADatabaseHiddenBehindAnAmbiguous550`, `testUnverifiable550IsPermissionDeniedNotFileNotFound`, `testRefusedLISTFallbackCannotProveAbsence`, `testLISTCannotProveAHiddenNameAbsent`.

KeeForge/Services/Cloud/FTPCloudProvider.swift
miquno: Fixed in 11de86b. The rev is now `sha256:<hex>` of the file bytes. Before the rename, the target is downloaded again and its hash is compared with `expectedRev`. `download` and `upload` report the hash of the bytes they transferred, not a re-read. The cost is one extra download per save and per `getMetadata`, noted under Limitations. Tests: `testUploadConflictsOnASameSecondSameSizeRemoteWrite`, `testDownloadRevDescribesTheTransferredBytesNotALaterWrite`.

KeeForge/Services/Cloud/FTPCloudProvider.swift
miquno: Fixed in 11de86b. Everything after `STOR` (stat, hash check, install) runs inside a catch that deletes the temporary upload before rethrowing. Uploads that outlive a dead connection carry their UTC creation time in the name, and a later save in the same folder deletes any older than 24 hours. The sweep needs a listing that shows dotfiles (MLSD does; vsftpd's LIST does not), which is noted under Limitations. Tests: `testFailedStatAfterTheTransferRemovesTheUpload`, `testFailedRenameRemovesTheUpload`, `testSaveSweepsOnlyStaleUploadLeftovers`.

PR #138, opened 2026-09-20
feat(cloud): open and save databases on FTP servers

miquno:
# Summary

Part of #115. Adds FTP as a storage provider, so a database can be opened, saved, and created directly on an FTP server instead of going through a Files-app integration. SFTP and FTPS are **not** included (see Limitations), which is why this does not close the issue.

`3e594c8` addresses the second review round: the lost set-aside reply, the open probe's deadline against a transferring `getMetadata`, and stale backups. Each reply on those threads names its test.

This replaces #136, which GitHub closed when its source fork was deleted. The commits are the ones already reviewed there — `38bbc5c` (feature) and `11de86b` (the four review fixes) — plus `5c2b584`, which silences a Swift 6 warning in the FTP test double. The review-fix commit is the one worth re-reading; the replies in #136 map each fix to its tests.

# Changes

- **Provider:** `FTPCloudProvider` plugs into the existing `CloudProvider` pipeline (cache, conflict gate, backups, conflict copies, pending AutoFill uploads) the same way WebDAV does. It is stateless and reads its credential from `CloudTokenStore` on each call.
- **Client:** `FTPClient` is built on Network.framework, so there is **no new dependency**.
  - Each operation opens its own logged-in session.
  - Data connections use passive mode: EPSV, falling back to PASV. The data connection always goes to the control host, never the address PASV advertises.
  - Listing and metadata use MLSD/MLST, falling back to LIST, SIZE and MDTM (vsftpd supports neither MLSD nor MLST).
  - Every read and write has a timeout, and task cancellation cancels the connection, so the 10 s open probe works.
  - Command arguments containing CR, LF or NUL are rejected.
- **Saves:** the upload goes to a hidden temporary name, the database is downloaded again and its hash compared with the recorded rev, then the upload is renamed over the database. The database is never written in place or deleted first.
  - A server that cannot rename onto an existing file (IIS): the database is renamed to a `.keeforge-backup` beside it, re-checked, and replaced. Any failure renames it back, over a fresh connection if the old one died.
  - Every failure after `STOR` deletes the temporary upload. Uploads that outlive a dead connection carry their creation time in the name and are swept by a later save in the same folder after 24 hours.
  - `createFile` refuses an existing name before and after the transfer.
- **Existence checks:** a 550 to `MLST` does not count as "missing", because RFC 3659 also allows it for denied access. `SIZE` and then the parent listing must show the name is absent; otherwise the operation fails as permission denied.
- **Rev:** `sha256:<hex>` of the file bytes. `download` and `upload` report the hash of the bytes they transferred, never a re-read.
- **Connect form:** `FTPConnectView`/`FTPConnectViewModel` take an `ftp://host[:port]/path` address, a username and a password. The path is relative to the login directory, RFC 1738 style. Plain FTP sends the login unencrypted, so connecting requires turning on **Allow Unencrypted FTP**, like WebDAV's HTTP opt-in. The browser picks the right form through a small `ManualConnectSheet`.
- **Reconnect:** the Reconnect action explains how to add the server again instead of reporting a configuration error.
- **Platforms:** FTP is available on iOS and macOS. Both targets already have outbound network access.
- **New Database:** the "Save To" picker is now a menu. With five destinations, the segmented control truncated "Local Files" on iPhone.
- **Strings and docs:** new strings are translated into all shipped locales. `NSLocalNetworkUsageDescription` now mentions FTP. Folder docs and CHANGELOG are updated.

# Testing

Run on `ff6874b`, against `main` at `7f60353`:

- `xcodebuild test -scheme KeeForge -destination 'platform=iOS Simulator,name=iPhone 17 Pro'` with `-only-testing:` for `FTPCloudProviderTests`, `FTPListingParserTests`, `FTPURLTests`, `FTPConnectViewModelTests`, `CloudProviderRegistryTests`, `CloudFileBrowserViewModelTests`, `CloudSyncCoordinatorTests`, `CloudDatabaseSaverTests` and `LocalizationTests`: **160 passed, 0 failed** (`FTPCloudProviderTests` alone: 53).
- Every test added for a review round was also run against the code without that fix, and each one fails there — including the backup-sweep case, which deletes the original outright.
- `xcodegen generate`: no project changes beyond the additive entries already committed.

What the tests cover: `FTPCloudProviderTests` drives the real protocol code against an in-memory FTP server (`Support/FakeFTPServer.swift`) — connect and persist, rejected login, missing folder, the opt-in check, MLSD listing, LIST+PASV against the control host, metadata with the SIZE/MDTM fallback, temp-then-rename upload, the post-transfer rev conflict, the IIS fallback (including a dropped connection and a lost reply), create-only refusals before and during the upload, 552/550 mapping, timeout reported as offline, connection refused, CR/LF injection, the reconnect message, temporary-upload cleanup and the stale-upload sweep.

Not run on this push:

- **`KeeForgeMacTests`.** Left to the `macos-unit-tests` check. The shared unit sources are the same ones the iOS run compiled, but a macOS-only compile or behavior difference would surface only in CI.
- **UI tests.** There is no UI test double for FTP, and the Mac UI was not exercised by hand.
- **Real servers.** The manual end-to-end checks against local vsftpd and pure-ftpd containers (create, list, metadata, download, upload with rev, stale-rev conflict, create conflict, wrong password, refused port, cancellation of an unanswered connect; and in the iOS app: add an FTP database, unlock, save, an external change raising the Save Conflict dialog, "Save as Conflict Copy", and creating a new database on FTP) were done against these commits when #136 was open. They were not repeated for this push, and the only change since is the test-double warning fix. IIS and PASV-only servers are covered only by the fake server.

# Limitations

- **No SFTP or FTPS.** SFTP needs an SSH library; the only maintained Swift option pulls in a personal fork of swift-nio-ssh, which needs a maintainer decision. Explicit FTPS (AUTH TLS) cannot be done on Network.framework.
- **Extra download per save and per open.** The rev is a content hash, so `getMetadata` and every save download the database once more. Because that probe scales with file size it is exempt from the open deadline (`metadataProbeTransfersContent`) and bounded by the client's own read timeouts instead. Cheap at KDBX sizes, but not free.
- **Last-moment writes.** A write that lands between the hash check and the rename is still lost on servers whose rename overwrites; FTP has no conditional write.
- **Sweep needs dotfiles.** Leftover uploads are only swept where the listing shows hidden files (MLSD does; vsftpd's `LIST` does not).
- **Backup can remain.** If the connection dies while the IIS path has the database set aside and something already stands at the original name, the `.keeforge-backup` is left in place rather than overwrite it. The delete after a successful install is retried over a fresh session, but a backup that survives that is never swept: nothing on the server distinguishes it from the copy above, so it is the user's to remove.
- **Create-only is not atomic.** FTP has no exclusive create, so a file appearing between the last existence check and the rename can be replaced on servers whose rename overwrites.
- **Absence rests on a parsed listing.** Where a server has neither `MLST`/`MLSD` nor `SIZE`, absence is decided from `LIST`, and the parser understands the Unix and IIS/DOS layouts only. A third layout parses as zero entries, which reads as an empty folder, so `createFile` could rename over an existing database on such a server. Every server that reaches that path in practice answers `SIZE`.
- **Permissions after save.** The rename puts a new file in place, so the saved database gets the server's permissions for new uploads (umask), not its previous mode.

# Checklist

- [ ] Updated `KDBXCompatibilityTests` if parser/writer/save behavior changed — not needed: no parser, writer or save-format change, only a new transport behind `CloudProvider`
- [ ] Ran a `KeeForgeMacTests` slice when shared or Mac behavior changed — not run locally; relying on `macos-unit-tests`


KeeForge/Services/Cloud/FTPCloudProvider.swift
```diff
@@ -0,0 +1,560 @@
...
+        in session: FTPSession,
+        location: FTPServerLocation,
+        credential: FTPCredential
+    ) async throws {
+        do {
+            try await session.rename(temporaryPath, to: path)
+            return
+        } catch FTPClientError.renameRefused {}
+
+        try await session.rename(path, to: backupPath)
```
crazytan: **[P1] Reconcile a lost set-aside reply**

`RNTO` can succeed server-side and then lose the control reply. Because this rename is outside the guarded block that invokes `restore`, that outcome exits with the live path missing and the original stranded under `backupPath`. Please reconcile this command on a fresh session or include it in the guarded restore path; the current dropped-connection test only covers the subsequent install rename.

KeeForge/Services/Cloud/FTPCloudProvider.swift
```diff
@@ -0,0 +1,560 @@
...
+
+    func getMetadata(accountId: String, fileId: String) async throws -> CloudFileMetadata {
+        let (location, credential) = try resolveContext(accountId: accountId)
+        let path = Self.remotePath(base: location.basePath, fileId: fileId)
+
+        return try await perform(context: .read, location: location, credential: credential) { session in
+            guard let entry = try await session.stat(path), !entry.isFolder else {
+                throw CloudProviderError.fileNotFound
+            }
+            let data = try await Self.retrieveExisting(path, in: session)
```
crazytan, line 163: **[P1] Avoid a full transfer inside the bounded metadata probe**

This turns `getMetadata` into a full `RETR`, but cached opens cancel metadata probes after 10 seconds. Any database that takes longer than that to transfer will be treated as offline on every open, so remote revisions are never downloaded and the stale cache is opened indefinitely. Keep the bounded probe lightweight or avoid applying the fixed metadata deadline when FTP needs a content download.

KeeForge/Services/Cloud/FTPCloudProvider.swift
```diff
@@ -0,0 +1,560 @@
...
+                backupPath,
+                to: path,
+                discarding: temporaryPath,
+                in: session,
+                location: location,
+                credential: credential
+            )
+            throw error
+        }
+        try? await session.delete(backupPath)
```
crazytan: **[P2] Recover stale backup files**

If this deletion fails, the error is discarded and the stale-file sweep never considers `.keeforge-backup` files. An IIS server with transient or persistent delete failures can accumulate complete historical database copies after successful saves. Please retry or safely sweep aged backups once the live target is verified.

crazytan (changes requested):
Three remaining correctness and recovery issues are called out inline. The focused FTP/cloud/coordinator/localization test slices pass on iOS and macOS.

KeeForge/Services/Cloud/FTPCloudProvider.swift
miquno: Fixed in 3e594c8. The set-aside rename now runs inside the block that restores on failure, so an `RNTO` carried out with its reply lost is reconciled rather than leaving the database under the backup name: `restore` checks the live path on a fresh session when the old one died, and moves the backup back only if nothing already stands there. Test: `testLostSetAsideReplyPutsTheDatabaseBack`, which arms the drop on the set-aside's own `RNFR` so the refused install rename before it is not the one that loses its reply. It fails with the previous code, leaving the database stranded under the backup name.

KeeForge/Services/Cloud/FTPCloudProvider.swift
miquno: Fixed in 3e594c8, by the second route you named: the fixed deadline no longer applies to this probe. `CloudProvider.metadataProbeTransfersContent` (false by default, true for FTP) tells `syncIfNeededForOpen` that the probe scales with file size, and it then passes no deadline. The probe stays bounded by `FTPClient`'s own 30 s per-read and connect timeouts, so a black-holed server still falls back to the cache rather than hanging the unlock sheet; that is what the deadline was for.

I did look at keeping the probe lightweight instead — a cheap stamp+size rev with the strong hash in `contentHash`. It needs more than this PR should carry: `upload` would have to receive the recorded baseline hash to keep the check you asked for in #136, and a probe-only open writes `remoteContentHash = syncedMetadata.contentHash`, which would erase that baseline. Happy to do it as separate work if you would rather have the cheap probe.

Test: `testProbeThatTransfersContentIsNotCutOffByTheDeadline` — a probe slower than the deadline is not cancelled and the newer copy is downloaded. It fails with the previous code, which returned the stale cache as `.offlineCached`.

KeeForge/Services/Cloud/FTPCloudProvider.swift
miquno: Fixed in 3e594c8. The sweep now covers both scratch kinds, so a discarded delete is recovered by the next save in that folder rather than accumulating copies.

A backup goes only once the database it names is confirmed present. Writing that check against the path being saved would have been wrong: the folder holds backups of every database beside it, so a save of `personal.kdbx` would have swept away the backup of a different database whose install had died — its only remaining copy. The scratch name carries the database it belongs to, so the sweep stats that name and caches the answer per database.

Tests: `testSaveSweepsOnlyStaleScratchLeftovers` now has aged backups of two present databases swept, an aged backup of a database that is not there kept, and a recent backup kept. It fails with the previous code, which left all of them in place.

KeeForge/Services/Cloud/FTPCloudProvider.swift
```diff
@@ -0,0 +1,593 @@
...
+        var databaseExists: [String: Bool] = [:]
+        for entry in entries where !entry.isFolder {
+            guard let scratch = Self.scratch(named: entry.name), scratch.createdAt < cutoff else {
+                continue
+            }
+            if scratch.kind == .backup {
+                let base = scratch.base
+                if databaseExists[base] == nil {
+                    let live = Self.sibling(of: path, named: base)
+                    databaseExists[base] = ((try? await session.stat(live)) ?? nil) != nil
```
crazytan: **Do not discard the only original copy**

Existence at the live path is not sufficient proof that this backup is redundant. The state already modeled by `testUnrestorableOriginalSurvivesUnderItsBackupName` leaves the original only under this backup name; after 24 hours, creating or uploading any unrelated database at the normal path makes this check true and the sweep permanently deletes the original. The cleanup needs durable evidence that this backup came from a completed install, rather than merely checking that the path has since been occupied.

crazytan (dismissed):
The fixes address the previous review and the focused FTP/cloud coordinator suites pass on iOS and macOS. One remaining backup-cleanup concern is called out inline.

KeeForge/Services/Cloud/FTPCloudProvider.swift
miquno: Fixed in ff6874b — you are right, and the check was worse than unproven: I verified it deletes the original in exactly the state you describe.

Backups are out of the sweep again. The delete right after a successful install is the only point at which a backup is known to be redundant, so a refusal there is now retried over a fresh session (`discardBackup`) instead of being deferred to a sweep that cannot tell that case from the one `testUnrestorableOriginalSurvivesUnderItsBackupName` leaves behind. A backup that survives both is left in place, and the PR still lists that under Limitations. The sweep and the scratch-name parser are byte-identical to their state before the previous round; only uploads are swept, whose bytes the client still holds.

Tests:
- `testStaleBackupSurvivesADifferentDatabaseTakingItsName` builds your scenario — an aged backup as the only copy, a different database at the name, then a save. Against the previous code it fails with the backup gone (`nil` instead of its bytes).
- `testRefusedBackupDeletionIsRetriedOverAFreshConnection` refuses the `DELE` with a 450 and pins that the backup is removed over a second control connection. Against the previous code the backup stays and only one connection is opened.

crazytan (approved):

Issue #87, opened 2026-08-13
feat(files): offline access for network-hosted databases opened through the Files app


crazytan:
## Problem

From August feedback:

> I have my keypass file at home on my NAS. When away from home though, I would need a local copy on my device accessible and usable that syncs again when getting home. Currently this does not work so I have no access to my pws at all when outside my home lan.

KeeForge's two open paths are asymmetric about offline availability:

- **Cloud-provider databases** (Dropbox, OneDrive, WebDAV) keep a cached copy. `CloudSyncCoordinator` falls back to it when the remote is unreachable and surfaces `.offlineCached` ("Using the cached copy offline."), so these already work off-network.
- **Bookmark-backed databases opened through the Files app** — an SMB share to a NAS, or any other File Provider — have no cache at all. The list deliberately does not resolve or probe those bookmarks (`DatabaseListViewModel.reload`, because an offline SMB URL can block synchronously and freeze launch), and the open path simply fails once the provider is unreachable.

So a NAS-hosted database added through Files is completely unavailable away from the LAN, with no local copy and no read-only fallback.

A NAS reachable over WebDAV can be connected through KeeForge's own WebDAV provider instead, which does get the cache — but nothing in the app tells the user that, and it does not help SMB-only setups.

## Desired behavior

A database hosted on a network share stays readable when the share is unreachable, and re-synchronizes when it becomes reachable again, without weakening conflict detection.

## Proposed direction

- Evaluate extending the cloud cache model to bookmark-backed references, so a successful open records a local copy that later opens can fall back to.
- Keep the cached open explicitly labeled as stale, reusing the existing `.offlineCached` banner rather than inventing a second vocabulary.
- Preserve the launch-time invariant: never resolve or probe a File Provider bookmark while building the list. A fallback decision has to happen on the bounded open path.
- Never let a cached open silently overwrite the share on the next save — the existing open-time hash / revision gate must still run once the provider is reachable again, and a divergence must conflict.
- Consider surfacing WebDAV as the supported route for NAS databases where the device offers it, since that path already behaves correctly.

## Questions

- Should an offline cache for File Provider databases be opt-in per database? It puts a copy of the database in KeeForge's container that the user did not explicitly ask for, which is a meaningful change for someone who deliberately keeps their vault off the device.
- Should a cached-only open be read-only until the share returns, rather than accepting edits that cannot be pushed?

## Related

[...]

PR #140, opened 2026-09-21
fix(files): relink a Files database whose file was replaced

### LINKED ISSUE #53: fix(files): recover when a File Provider replaces or temporarily loses a vault
## Summary

Three July reports on v1.10.2–1.10.3 describe Files-backed databases becoming unavailable after external changes: two Nextcloud cases and one iCloud-hosted KDBX edited from KeePassXC on Mac. They later fail with `file.not_found` or require removal and re-addition.

This is related to #13, but the reports arrived after that fix and specifically involve File Provider replacement or availability transitions. Users also requested a safe local encrypted cache for transient provider outages.

## Suggested direction

- Detect when a bookmark no longer resolves to the current item at the expected path.
- Offer an explicit reselect/rebind flow.
- Evaluate an encrypted local cache with clear stale/conflict handling.
- Preserve protection against reading or re-bookmarking copies in Recently Deleted.

## Acceptance criteria

- Externally replacing or syncing the same KDBX path is recoverable.
- Temporary provider unavailability does not discard the last known encrypted copy.
- Cached bytes never overwrite a newer provider version without conflict detection.

miquno:
# Summary

Fixes #53

A database added through the Files app becomes unusable after its File Provider (Nextcloud, iCloud Drive) replaces the synced KDBX, or after a Files-app Replace. Unlock fails with `file.not_found` or `file.in_recently_deleted`, and the only way back was to remove the database and add it again. Removal also deletes its cached copy, backups, saved biometric key, and AutoFill identities.

**Expected:** an externally replaced or synced file at the same place is recoverable without losing the database's settings or its last cached copy.

**Root cause:** a Files database is stored only as a security-scoped bookmark, which follows the provider's item identity rather than the path. Once the provider swaps the item, the bookmark either stops resolving (`DatabaseListStore.locateDatabaseFile` returns nil, so `file.not_found`) or follows the old copy into Recently Deleted. Nothing let the existing reference point at the current file. "Choose Different File" only goes back to the list, and adding the file again creates a second, unrelated reference.

# Changes

- `DatabaseListStore.relinkLocalDatabase(id:to:)` points an existing bookmarked reference at a picked file. It changes only `bookmarkData`, `filename` and `isDocumentsResident`, so id, nickname, key file, Keychain key, AutoFill settings and backups are kept. It reseeds the shared cache from the picked file. It refuses a pick in Recently Deleted and a file another listed database already uses (the existing `duplicateFile` error). It returns nil for cloud-backed references and for KeeForge-only references without a bookmark, because their cache *is* the database and reseeding it would overwrite it.
- The unlock failure screen offers **Locate Database File** for `file.not_found`, `file.in_recently_deleted`, `file.permission_denied` and `file.read_failed` on bookmarked local databases. It is not offered for `file.read_timeout`, where the server is unreachable and another file would not help. After a relink the screen returns to the unlock form. A relink marks the reference `hasUnverifiedRelink` (persisted), and `markDatabaseOpened` clears it on the next successful unlock. While it is set, every failure except a read timeout keeps offering the action, so picking a wrong but valid KDBX (which fails as `auth.invalid_credentials`) can be corrected. A plain wrong password on a verified reference still does not offer it.
- `UnlockView`'s key-file and database pickers share one `.fileImporter` through the existing `PickerPresentationState`, so the completion still knows which picker ran after dismissal. The database-selection check moved from `DatabaseListView` to `DocumentPickerService.isSupportedDatabaseSelection(_:)` so both screens use it.
- The Recently Deleted summary and the `LocalDatabaseFileError.databaseInTrash` text, which the relink picker shows, now point to Locate Database File instead of remove-and-add (app and AutoFill catalogs). It has a new button label, both translated for all shipped locales, and the catalog is normalized.
- CHANGELOG entry and notes in the Persistence and Unlock folder docs.

Out of scope, per the discussion on the issue: no offline fallback to the cached copy while a provider is unreachable. That is still #87. So cached bytes are never used to open a database or written to the provider. The relink only ever copies provider → cache, and the open-time SHA-512 save gate is unchanged.

# Testing

- `xcodebuild test -scheme KeeForge` on an iPhone 17 simulator with `-only-testing:` for `KeeForgeTests/DatabaseListStoreTests`, `DatabaseViewModelTests`, `DatabaseListViewModelTests` and `LocalizationTests`: **286 passed, 0 failed** (run after the final change).
  - New in `DatabaseListStoreTests`: relink keeps identity and reseeds the cache; refuses a trashed pick; refuses a file another reference uses; allows relinking to its own file; leaves a KeeForge-only database and its cache untouched; ignores an unlisted id.
  - New in `DatabaseViewModelTests`: the issue's scenario, where an unresolvable bookmark fails with `file.not_found`, then relink, then unlock succeeds under the same id and nickname. Also covers the action being offered for Recently Deleted but not for read timeouts or wrong passwords, and staying offered after a wrong-file relink (also from a fresh view model) until an unlock succeeds.
  - Removing the trash check, the duplicate check, or the no-bookmark guard from `relinkLocalDatabase` makes the matching test fail (checked by temporarily removing each).
- `swift scripts/normalize-xcstrings.swift --check`: passes.
- `xcodebuild build -scheme KeeForgeMac -destination 'platform=macOS' CODE_SIGNING_ALLOWED=NO`: succeeds.
- Manual, iPhone 17 simulator: a Documents database whose file is gone shows `file.not_found` with the new button. Picking a file already in the list shows the "already in your database list" alert. Picking a copy in a subfolder returns to the unlock form, and the same entry unlocks.

Not tested:
- A real Nextcloud or iCloud Drive item replacement on a device. The simulator can't reproduce a provider swapping item identity, so the tests use an unresolvable bookmark, which is the state that replacement leaves behind.
- `KeeForgeMacTests` was not run locally; the `macos-unit-tests` check covers it. The new tests are platform-neutral.

# Checklist

- [x] Ran `xcodegen generate` after adding/removing/moving source files (no files added or moved)
- [x] Updated the nearest folder-local `README.md` if the folder's map changed (Persistence and Unlock `AGENTS.md`)
- [x] Kept AutoFill extension target membership/imports in sync for shared code (`DocumentPickerService` addition is Foundation-only)
- [ ] Updated `KDBXCompatibilityTests` if parser/writer/save behavior changed (not applicable, no parser/writer/save change)
- [x] Preserved accessibility identifiers or updated the affected UI tests (new: `unlock.relink`)
- [ ] Ran a `KeeForgeMacTests` slice when shared or Mac behavior changed (left to CI; Mac app target build verified locally)


KeeForge/ViewModels/DatabaseViewModel.swift
```diff
@@ -702,6 +702,40 @@ final class DatabaseViewModel {
...
+    /// Whether the failure screen may offer to relink this database to a file
+    /// the user picks. Read timeouts are left out: the server is unreachable,
+    /// and a different file would not bring it back.
+    var canRelinkDatabaseFile: Bool {
+        guard let failure = openFailure,
+              databaseReference.isCloudBacked == false,
+              databaseReference.bookmarkData != nil else {
+            return false
+        }
+        return Self.relinkableFailureCodes.contains(failure.errorCode)
```
crazytan: After a relink, please preserve a way to relink again until this reference successfully unlocks. If the user picks the wrong valid KDBX, the next failure is `auth.invalid_credentials`; this whitelist then hides Locate Database File even though the bookmark and cache have already moved to the selected file. Tracking an unverified relink until successful unlock would keep ordinary wrong-password flows unchanged while avoiding a return to remove-and-add.

KeeForge/Views/Unlock/UnlockView.swift
```diff
@@ -481,6 +502,34 @@ struct UnlockView: View {
...
+        switch result {
+        case .success(let url):
+            guard DocumentPickerService.isSupportedDatabaseSelection(url) else {
+                selectionAlert = DocumentPickerService.invalidDatabaseSelectionAlert()
+                return
+            }
+            do {
+                try viewModel.relinkDatabaseFile(to: url)
+            } catch {
+                selectionAlert = DocumentPickerService.pickerFailureAlert(for: error)
```
crazytan, line 526: This text is now surfaced directly by the relink picker when the selected file is in Recently Deleted, but it still recommends remove-and-add—the destructive workflow this PR replaces. Please update the guidance (and the shared app/AutoFill catalog entries) to point back to Locate Database File.

crazytan (dismissed):
Approved. I left two inline follow-ups for recovery-state resilience and the remaining Recently Deleted guidance.

PR #141, opened 2026-09-21
fix(search): refresh a result's TOTP indicator after adding or removing a code

### LINKED ISSUE #111: Indicator icon after adding / removing TOTP
### What happened?

Right after adding TOTP, the search list containing the target entry (that just got TOTP added) doesn't display the green/TOTP icon, until after the search is cleared and re-entered.

### Steps to reproduce

Search for target entry that needs TOTP
Open entry and add TOTP
Tap Back
(The entry doesn't reflect green TOTP icon)
Tap Cancel to clear search
Re-search the entry name
(The entry now reflects green TOTP icon)

NOTE:  The same issue occurs in reverse, i.e. after removing TOTP from an entry, the green TOTP icon still appears in search until the search is cleared and then re-searched.

### KeeForge version

1.14.0 and 1.15.0 (beta)

### iOS version and device

18.7.2 iPhone 13 and 26.6.1 iPhone SE 2020

### Where does the bug occur?

Editing / saving

_No response_

### Anything else?

Image

miquno:
# Summary

Fixes #111

Adding a verification code to an entry opened from the search results, then going back, left the result row without the green code icon. Removing a code did the opposite: the icon stayed. Either way it only corrected itself after clearing the search and searching again.

**Expected:** the result row shows the entry's current code state as soon as you go back.

**Root cause:** `KPEntry` compares equal by ID alone (`NavigationConformances.swift`). After an edit that keeps the same matches, `rebuildDerivedState` reassigns `DatabaseViewModel.searchResults` with entries that are `==` to the old ones. The `@Observable` setter skips notifying for an equal `Equatable` value, so `SearchView` never passes the new array to `EntryListView`. `EntryListView` still re-renders on `contentRevision`, but it draws its stale `entries`. The group list is not affected because it derives its entries from the view model inside its own body.

# Changes

- `EntryListView` resolves each row through `DatabaseViewModel.entry(withID:)`, which reads the rebuilt index, and falls back to the passed entry.
- I tried keying the whole search results subtree on `.id(viewModel.contentRevision)` first and dropped it. Every successful save reassigns `rootGroup` and bumps `contentRevision`, so that would recreate `EntryListView` and dismiss a Duplicate or Move sheet opened while a save is still finishing. The row-level lookup leaves those `@State` hosts alone.
- Two UI tests in `TOTPEnrollmentUITests` cover the issue's steps in both directions (add a code to "Union Bank", remove "Union News"'s).
- CHANGELOG entry, a note on the ID-only equality in `KeeForge/Extensions/AGENTS.md`, and the test list in `KeeForgeUITests/README.md`.

The tag browser on iPhone and iPad draws the same `EntryListView`, so it gets this fix too, but no test covers it there. macOS search and the tag browser use `MacEntriesList` / `MacEntryRow`, which passed entries the same way. Since review, `MacEntryRow` also resolves its entry through `DatabaseViewModel.entry(withID:)`, which covers both Mac lists and the group column. No automated test covers the Mac path: `KeeForgeMacUITests` does not run in CI and I don't run Mac tests locally. The Mac app target builds.

# Testing

All runs on an iPhone 17 simulator (iOS 27.0):

- Before the fix, `-only-testing:KeeForgeUITests/TOTPEnrollmentUITests/testAddingCodeFromSearchResultShowsIndicatorAfterGoingBack` and `.../testRemovingCodeFromSearchResultHidesIndicatorAfterGoingBack`: **both failed** at the after-going-back assertion. The starting-state checks passed (no icon on Union Bank, icon on Union News).
- Same two tests on the final code: **2 passed**.
- `-only-testing:` `KeeForgeUITests/TOTPEnrollmentUITests`, `SearchResultsDeleteUITests`, `SearchResultsMoveUITests`, `UnlockedDatabaseSearchAndSortUITests`, `TagBrowserUITests`, and `EntryEditSmokeUITests/testEditedTitleAndUsernameRefreshGroupSearchAndTitleSorting`: **19 passed, 0 failed**. After that run I only tightened a test helper and edited a comment, and then re-ran the two regression tests plus `SearchResultsMoveUITests`: passed.
- `xcodebuild build -scheme KeeForgeMac -destination 'platform=macOS' CODE_SIGNING_ALLOWED=NO`: succeeds.

Not tested:
- `KeeForgeMacTests` was not run locally. The `macos-unit-tests` check covers it, and no unit sources changed.
- A physical device. The issue's iOS 18.7 / iPhone 13 setup was not reproduced; the simulator runs iOS 27.

# Checklist

- [x] Ran `xcodegen generate` after adding/removing/moving source files (no files added or moved)
- [x] Updated the nearest folder-local `README.md` if the folder's map changed (`KeeForgeUITests/README.md`, `KeeForge/Extensions/AGENTS.md`)
- [ ] Localized new UI strings in all shipped locales listed in AGENTS.md and ran `LocalizationTests` (no new strings)
- [x] Kept AutoFill extension target membership/imports in sync for shared code (no membership change)
- [ ] Updated `KDBXCompatibilityTests` if parser/writer/save behavior changed (not changed)
- [ ] Ran a `KeeForgeMacTests` slice when shared or Mac behavior changed (left to the `macos-unit-tests` check)


KeeForge/Views/Entry/EntryListView.swift
```diff
@@ -26,7 +26,10 @@ struct EntryListView: View {
                 ContentUnavailableView.search
             } else {
                 List(entries) { entry in
-                    entryRow(for: entry)
+                    // `KPEntry` equality is by ID, so after an edit that keeps
+                    // the same matches (adding a TOTP code) Observation treats
+                    // the caller's array as unchanged and `entries` is stale.
+                    entryRow(for: viewModel.entry(withID: entry.id) ?? entry)
```
crazytan, line 32: Non-blocking: `SearchResultsList` routes macOS through `MacEntriesList`, so this lookup does not cover the Mac path. `MacEntriesList` still passes its captured `entry` directly to `MacEntryRow`, leaving the same stale TOTP indicator possible there. Please consider applying the same current-entry resolution in the Mac counterpart, ideally with focused coverage.

crazytan (dismissed):
Approved. The iOS fix is sound and the two focused regression tests passed locally. I left one non-blocking inline note about applying the same refresh behavior to the macOS search list.

PR #142, opened 2026-09-21
feat(unlock): open databases that need a YubiKey challenge-response

### LINKED ISSUE #62: feat(unlock): support YubiKey HMAC-SHA1 challenge-response
## Problem

A July reporter could not open a KDBX 4.1 database that requires YubiKey HMAC-SHA1 challenge-response. KeeForge currently treats the missing challenge-response component like invalid credentials.

## Desired behavior

Support KeePass-compatible HMAC-SHA1 challenge-response using a connected YubiKey as part of the composite key.

## Considerations

- NFC and supported wired YubiKey transports on iOS/iPadOS.
- Correct KDBX challenge construction and slot selection.
- Cancellation, retry, and device-removal behavior.
- Clear unsupported-method diagnostics instead of `auth.invalid_credentials`.
- Keychain/biometric Quick Launch must not persist or emulate the hardware response.

miquno:
# Summary

Closes #62

KeeForge couldn't open a database whose composite key includes a YubiKey HMAC-SHA1 challenge-response. The missing component was reported as `auth.invalid_credentials`. This PR adds KeePassXC-compatible challenge-response unlock on iPhone, and on iPad models with a Lightning port, over NFC or Lightning.

It is deliberately the first slice. **Sessions unlocked with a YubiKey are read-only.** `KDBXWriter` rotates the KDF salt on every save, so a save needs a fresh challenge-response on every save path (local, cloud, merge, conflict copy). That is a separate change that can build on this one.

# Changes

- **Key derivation** (`Models/HardwareKey.swift`), following KeePassXC's `CompositeKey::rawKey(transformSeed)` / `transform`:
  - The challenge is the KDBX 4 KDF salt (Argon2 salt or AES-KDF seed), PKCS#7-padded to 64 bytes (`YubiKeyInterfaceUSB::challenge`).
  - The key fed to the KDF becomes `SHA256(preKey || SHA256(response))`.
  - `KDBXCrypto.compositeKey(password:keyFileData:)` is split so the pre-key is reachable. Its output is unchanged, and a test pins that.
  - The parser and writer are untouched.
  - KDBX 3.1 files get `hardware_key.unsupported_database`, because KeePassXC folds the response in after the KDF there. KDBX 4 AES-KDF is supported: KeePassXC writes it under the legacy `c9d9f39a…` UUID and reads it back as its KDBX 4 AES-KDF (`KeePass2::kdfFromParameters`), which takes the response before the KDF like Argon2. `challenge-response-aeskdf.kdbx` pins it.
- **YubiKit 4.7.0** (Apache-2.0, SPM) talks to the key through `Services/Security/HardwareKeyService.swift`.
  - It is linked into the iOS `KeeForge` target only. The Mac app and both AutoFill extensions don't link it.
  - USB-C YubiKeys are not reachable: YubiKit's smart-card connection doesn't carry the OTP application.
  - Added to Acknowledgments and to the dependency list in CONTRIBUTING (all locales).
- **Unlock UI**
  - A "Hardware Key" picker (NFC or Lightning, slot 1 or 2) appears when the device has a transport. The choice is stored on `DatabaseReference.hardwareKey`.
  - A YubiKey alone counts as a key component.
  - While the app waits for the key, the opening screen shows Cancel.
- **Failure handling**
  - Cancel, timeout, key removed, slot not set up, unavailable, and communication errors become retryable `hardware_key.*` failures. They don't count toward the lockout.
  - A wrong YubiKey stays `auth.invalid_credentials`, with a summary that names the YubiKey.
  - Where the device could use a YubiKey but none is configured, the invalid-credentials summary now suggests the Hardware Key option. That is the case from the issue.
- **Lock during a pending request:** `lock()` cancels the YubiKey request. A Lightning request has no timeout.
- **Unlock attempt token:** each unlock attempt holds a token from `prepareForUnlock()`; neither its success nor its failure touches the state unless `finishUnlockAttempt` still accepts it. `lock()` drops the token, and backgrounding locks a still-running attempt whenever it would lock an unlocked session (iOS `lockOnBackground`, always on the Mac), so a late answer cannot unlock behind the lock. The token is separate from `lockCycleID`, which paces biometric auto-unlock.
- **Cancel wins:** the service re-checks cancellation after the YubiKit continuation resumes, and the view model does the same after the response task, so an answer already queued when Cancel was tapped is discarded.
- **Quick Launch / Keychain**
  - For hardware-key databases the Keychain item holds the pre-key, never the composite key. The response is never persisted, and biometric unlock still asks the YubiKey.
  - `DatabaseListStore.setHardwareKey` deletes the item whenever the hardware key is turned on or off, because the item's meaning changes.
- **AutoFill:** a hardware-key database is refused with an explanation, and biometrics is not offered for it.
- **Read-only:** the toolbar indicator and the Database Details footer explain why. `ReadOnlyIndicator` now takes the explanation from the view model.
- **Entitlements and Info.plist:**
  - `com.apple.developer.nfc.readersession.formats` (`TAG`)
  - The YubiKey OTP AIDs under `com.apple.developer.nfc.readersession.iso7816.select-identifiers`
  - `NFCReaderUsageDescription` (localized)
  - `com.yubico.ylp` under `UISupportedExternalAccessoryProtocols`
- **Fixture:** `TestFixtures/challenge-response.kdbx` is authored with pykeepass, which doesn't know about hardware keys. The generator derives the transformed key itself from the KeePassXC formula and an emulated YubiKey. The file opens only with password and response.
- Strings are translated in all shipped locales. CHANGELOG, fixture README, and folder docs are updated.

# Testing

iOS Simulator (iPhone 17 Pro, iOS 27.0, Xcode 27.0), all passed:

- `ChallengeResponseKeyTests` (new, 10 tests):
  - padded-salt challenge
  - KDBX 3.1 refusal, and the AES-KDF fixture opening with password + emulated key
  - composite-key vectors computed separately in Python
  - fixture opens with password + emulated key, and is rejected with password alone or with a different key
- `HardwareKeyUnlockTests` (new, 16 tests), with an emulated YubiKey injected in place of YubiKit:
  - read-only session, and the refused save
  - lock ends the session
  - wrong key, and the missing-key hints
  - cancel, lock while waiting, disconnect
  - a key that answers after lock, after a locking background, or after Cancel does not unlock (fail with the old code); backgrounding with lock-on-background off lets the unlock finish
  - KDBX 3.1 refused before the key is asked
  - biometric unlock with the stored pre-key
  - Keychain item dropped when the key is turned on or off, and kept on a slot change
- `CredentialProviderCoordinatorTests` (new test: hardware-key database refused with an explanation, no biometric option)
- `DatabaseReferenceTests` (Codable round trip and legacy decode)
- Existing suites, re-run after the last code change, 491 tests total with no failures or skips:
  - `DatabaseViewModelTests`
  - `KDBXParserTests`
  - `KDBXWriterTests`
  - `DatabaseListStoreTests`
  - `CredentialProviderSaveTests`
  - `KeychainServiceTests`
  - `KeyFileProcessorTests`
  - `LocalizationTests`
- `swift scripts/normalize-xcstrings.swift --check`: clean
- `python3 TestFixtures/generate_fixtures.py challenge-response --check`: verified

macOS: `KeeForgeMac` build and build-for-testing succeed (`CODE_SIGNING_ALLOWED=NO`), so the shared sources and the new tests compile for the Mac. The `KeeForgeMacTests` run is left to the `macos-unit-tests` job.

**Not tested, and why:**

- No physical YubiKey or device was available. The YubiKit path in `HardwareKeyService` has not run against a real key: the NFC and Lightning sessions, the error-code mapping, the system NFC sheet, and a touch timeout. The simulator has no NFC, so the Hardware Key row doesn't appear there, which is also why there are no screenshots.
- Compatibility with a real KeePassXC database is shown through the formula and the independently derived fixture, not through a file written by KeePassXC. `keepassxc-cli` needs a real key.

Manual check with an NFC-capable iPhone and a YubiKey 5 NFC:

1. Program slot 2: `ykman otp chalresp --touch --generate 2`
2. In KeePassXC, add the YubiKey challenge-response to a KDBX 4 database (Argon2 or AES-KDF).
3. In KeeForge, choose "YubiKey via NFC, Slot 2", enter the password, unlock, and hold the key to the phone. The database should open read-only.
4. Also try: dismiss the NFC sheet, pick slot 1, and pull the key away mid-read.

# Before release (maintainer side)

- **NFC:** the App ID needs the "NFC Tag Reading" capability. Otherwise signing with the new entitlement fails.
- **Lightning:** declaring `com.yubico.ylp` needs Yubico's MFi product-plan registration for App Review. If that isn't wanted, dropping `UISupportedExternalAccessoryProtocols` leaves an NFC-only build; the Lightning option then doesn't appear.

# Checklist

- [x] Ran `xcodegen generate` after adding/removing/moving source files (only the additive project lines are committed)
- [ ] Updated `KDBXCompatibilityTests` if parser/writer/save behavior changed: not applicable, since parser, writer, and save paths are unchanged
- [ ] Ran a `KeeForgeMacTests` slice when shared or Mac behavior changed: compiled only, the run is left to CI


KeeForge/ViewModels/DatabaseViewModel.swift
```diff
@@ -788,7 +842,21 @@ final class DatabaseViewModel {
             cloudSyncStatus = readResult.cloudSyncStatus
             try cacheDatabaseCopyForLocalDatabase(data)
 
-            let compositeKey = try KDBXCrypto.compositeKey(password: password, keyFileData: keyFileData)
+            let hardwareKey = databaseReference.hardwareKey
+            let compositeKey: SymmetricKey
+            let quickLaunchKey: SymmetricKey
+            if let hardwareKey {
+                let preKey = try KDBXCrypto.preKey(password: password, keyFileData: keyFileData)
+                compositeKey = try await hardwareKeyCompositeKey(
```
crazytan, line 898: **[P1] Invalidate pending unlocks across lifecycle changes**

A hardware-key unlock can remain in `.unlocking` while the app backgrounds, but `handleSceneDidEnterBackground()` returns unless the state is already `.unlocked`. The outer unlock later calls `finalizeSuccessfulUnlock` without checking whether the lock cycle changed. I reproduced this with `lockOnBackground = true`: pausing the hardware response, backgrounding, and then returning a valid response left the view model unlocked with a populated `rootGroup`. Please bind the attempt to an epoch/token, invalidate it on background, lock, and cancel, and verify it before both success and error state mutations.

KeeForge/Models/HardwareKey.swift
```diff
@@ -0,0 +1,79 @@
...
+        guard case .kdbx4 = try KDBXParser.parseFileVersion(from: data) else {
+            throw HardwareKeyError.unsupportedDatabase
+        }
+        var reader = DataReader(data: data)
+        _ = try KDBXParser.parseVersion(from: &reader)
+        let header = try KDBXParser.parseHeader(&reader)
+
+        // KeePassXC folds the response in before the KDF only for these; a
+        // KDBX 4 file with the legacy AES-KDF UUID never carries one.
+        guard let kdfUUID = header.kdfParameters["$UUID"] as? Data,
```
crazytan, line 52: **[P2] Accept the KDBX4 AES-KDF challenge-response case**

KeePassXC also supports challenge-response with the KDBX4 AES-KDF: the response is incorporated into the raw composite key before the selected KDF is applied. Restricting this guard to Argon2d/Argon2id rejects valid KeePassXC databases, and the new refusal test locks in that incompatibility. Please allow the AES UUID for KDBX4 and cover it with an interoperable fixture; the existing file-version guard can continue rejecting KDBX3.1. References: [CompositeKey.cpp](https://github.com/keepassxreboot/keepassxc/blob/develop/src/keys/CompositeKey.cpp), [KeePass2.cpp](https://github.com/keepassxreboot/keepassxc/blob/develop/src/format/KeePass2.cpp).

KeeForge/Services/Security/HardwareKeyService.swift
```diff
@@ -0,0 +1,201 @@
...
+                case .nfc:
+                    YubiKitExternalLocalization.nfcScanAlertMessage = String(
+                        localized: "Hold your YubiKey near the top of your device."
+                    )
+                    YubiKitManager.shared.startNFCConnection()
+                case .lightning:
+                    YubiKitManager.shared.startAccessoryConnection()
+                }
+            }
+        } onCancel: {
```
crazytan, line 88: **[P2] Make cancellation win over a queued success callback**

The cancellation handler and the YubiKit callback enqueue independent main-actor `finish` calls. If the success task reaches `finish` first, the continuation returns normally even though the surrounding task is cancelled, because there is no cancellation check after the continuation resumes. Please store the continuation result, call `Task.checkCancellation()`, and only then return it. The view-model attempt token should remain the final guard against late completion restoring a locked session.

crazytan (changes requested):

KeeForge/Services/Security/HardwareKeyService.swift
```diff
@@ -0,0 +1,209 @@
...
+}
+
+extension YubiKeyConnection: YKFManagerDelegate {
+    nonisolated func didConnectNFC(_ connection: YKFNFCConnection) {
+        nonisolated(unsafe) let connection = connection
+        Task { @MainActor in self.sendChallenge(over: connection, transport: .nfc) }
+    }
+
+    nonisolated func didDisconnectNFC(_ connection: YKFNFCConnection, error: Error?) {
+        Task { @MainActor in self.finishActiveRequest(transport: .nfc, with: .disconnected) }
```
crazytan, line 192: [P2] When the NFC sheet is cancelled or times out after a tag connects, YubiKit can deliver that CoreNFC error through didDisconnectNFC. This always reports `disconnected`, so the user sees a removed-key message instead of cancellation or timeout. Please classify the supplied error and reserve `disconnected` for a lost connection.

crazytan (approved):
Approved with one P2 follow-up noted inline.

PR #143, opened 2026-09-21
fix(cloud): let cloud browsers list databases without a .kdbx extension

### LINKED ISSUE #113: fix(cloud): databases whose filename does not end in .kdbx cannot be selected from cloud browsers
## Problem

Reported in August feedback, as a follow-up from the NAS/offline thread that became #87: a user who connected their Nextcloud over WebDAV found their database was not selectable, because they deliberately store it under a filename that does not end in `.kdbx` for obscurity. They asked for either an unfiltered listing or an option to allow other extensions.

The local and cloud paths disagree about what counts as a selectable database.

**Local (Files app) — extension-agnostic.** `DocumentPickerService` offers `[databaseContentType, .item]`, so the picker will surface any file, and the service validates the selection by KDBX magic bytes (`kdbxMagic = 03 D9 A2 9A 67 FB 4B B5`) rather than by name. A vault called `notes.bin` opens fine.

**Cloud browsers — hard-filtered on the filename.** All three providers drop anything that does not end in `.kdbx`, so such a file is not merely un-preselected, it is invisible in the browser:

- `KeeForge/Services/Cloud/WebDAVCloudProvider.swift:418`
- `KeeForge/Services/Cloud/DropboxCloudProvider.swift:750` (plus a server-side `fileExtensions: ["kdbx"]` search filter at :472)
- `KeeForge/Services/Cloud/OneDriveCloudProvider.swift:1126`

So the same database is selectable through Files but unreachable through KeeForge's own WebDAV client — which is the route #87 recommends for exactly this NAS use case.

Renaming a vault away from `.kdbx` is a documented habit among KeePass users, and other clients (KeePassXC, Strongbox, KeePassium) do not require the extension.

## Desired behavior

A KDBX database stored on a cloud/WebDAV server can be selected regardless of its filename extension, without turning the browser into an undifferentiated file listing.

## Proposed direction

- Keep `.kdbx` as the default filter so the common case stays clean, and add an explicit "Show all files" affordance in the cloud browser rather than removing the filter outright.
- Validate the selection the way the local path already does — read the header and check the KDBX magic — so a wrong pick fails with a clear "this isn't a KeePass database" message instead of a parse error.
- Note the cost: the local path can sniff bytes cheaply, but a cloud browser would have to download a candidate before it can validate. Consider a range request for the first 8 bytes where the provider supports it (WebDAV and OneDrive do; Dropbox does not), and fall back to validating on open.
- Dropbox needs care beyond the client-side filter: `fileExtensions: ["kdbx"]` is passed to Dropbox's search API, so unfiltered browsing has to avoid that search path entirely.

## Related

- #87 — offline access for network-hosted databases; the same reporter moved to WebDAV on that advice and hit this next.

miquno:
# Summary

Fixes #113

A KDBX database stored on Dropbox, OneDrive, WebDAV, or FTP under a name that does not end in `.kdbx` could not be selected: the cloud browser never listed it. The same file opens fine through Files, which checks the KDBX magic instead of the name.

**Expected:** such a database is selectable from the cloud browser, while the browser still shows only `.kdbx` files by default.

**Root cause:** each provider's listing dropped every file whose name does not end in `.kdbx` (`WebDAVCloudProvider`, `FTPCloudProvider`, `DropboxCloudProvider`, `OneDriveCloudProvider`), and Dropbox search additionally sent `fileExtensions: ["kdbx"]` to the server. Nothing let the user get past the filter.

**Fix:** `CloudProvider.listFiles` takes `includesAllFiles`. Every provider filters through one rule, `CloudFile.isListed` (folders always, `.kdbx` files by default, everything when asked). Dropbox search drops its server-side extension filter only when all files are requested. The file browser gets a "Show All Files" switch at the top of the list; it carries into subfolders. The empty state ("No Databases") also offers a Show All Files button.

Picking a file that is not a KeePass database fails on unlock through the existing parser errors: "Not a KeePass Database" (`format.invalid_signature`), or "File is truncated" for files under 8 bytes. The issue's optional range-request pre-check (reading the first 8 bytes before selection) is not part of this change.

# Changes

- `CloudSyncModels.swift`: `CloudFile.isListed(name:isFolder:includesAllFiles:)`, the single listing rule.
- `CloudProvider.swift` and all four providers plus both UI-test doubles: `listFiles(…, includesAllFiles:)`.
- `DropboxCloudProvider`: `searchOptions(path:includesAllFiles:)` omits `fileExtensions` when listing all files. The search pagination loop now checks for cancellation, because an unfiltered search can span many pages and every keystroke starts a new one.
- `FTPCloudProvider`: the provider's own `.keeforge-upload` / `.keeforge-backup` scratch files are never listed, even with Show All Files. Only names in the exact generated shape `.<name>.<UTC time>-<8 hex>.<kind>` count; the listing and the stale-upload sweep share one parser, so a user's `.vault.keeforge-backup` stays visible (negative tests added). The provider already relied on them staying out of listings, and an upload leftover opened as a database could later be deleted by the stale-upload sweep. Other dotfiles are listed when all files are shown.
- `createFile` on Dropbox and OneDrive maps the created item with `includesAllFiles: true`. A conflict copy keeps the database's own extension (`vault (conflict …).bin`). With the old filter, that upload would have succeeded and then failed with "upload did not return a file", which this fix would otherwise have made reachable.
- `OneDriveDriveItem` and its facet types are now internal so the mapping can be unit-tested.
- `CloudFolderBrowserViewModel`: the flag is part of `requestKey`, so toggling reloads the listing. The destination-folder picker passes `false` (unchanged behavior).
- `CloudFileBrowserView`: the Show All Files switch and empty-state button. The file row body moved into a `fileRow(_:)` helper so the list can hold the switch section; the row itself is unchanged.
- New strings "Show All Files" localized in de, fr, es, ja, zh-Hans, zh-Hant; catalog normalized.
- Docs: `CHANGELOG.md` (Unreleased → Fixes), `Services/Cloud/AGENTS.md`, `KeeForgeUITests/README.md`.

# Testing

Local, iPhone 17 Pro simulator (iOS 26.5):

- `xcodebuild test … -only-testing:KeeForgeTests/CloudFileBrowserViewModelTests -only-testing:KeeForgeTests/WebDAVCloudProviderTests -only-testing:KeeForgeTests/FTPCloudProviderTests -only-testing:KeeForgeTests/FTPListingParserTests -only-testing:KeeForgeTests/DropboxCloudProviderTests -only-testing:KeeForgeTests/OneDriveCloudProviderTests -only-testing:KeeForgeTests/CloudSyncCoordinatorTests -only-testing:KeeForgeTests/LocalizationTests -only-testing:KeeForgeTests/DatabaseViewModelTests/testCloudUnlockTransitionsToErrorWhenDownloadedDataCannotBeParsed`: **220 passed, 0 failed** (on the final commit).
- `FTPCloudProviderTests/testShowingAllFilesStillHidesTheProvidersOwnScratchFiles` failed before the scratch-file fix (it listed both scratch files) and passes after it.
- New regression tests fail against the old rule: with `CloudFile.isListed` and the Dropbox search options temporarily restored to `.kdbx`-only, the WebDAV, FTP, Dropbox listing, Dropbox search, and OneDrive tests failed (5 failures), then passed again with the fix.
- `KeeForgeUITests/WebDAVShowAllFilesUITests` (new, end to end): `vault.bin` is hidden by default, appears after Show All Files, then opens and unlocks. Ran with `-test-iterations 3`: **3/3 passed**; on the final commit, again with `WebDAVAddFlowUITests`, `-test-iterations 2`: **4/4 passed**.
- Existing cloud UI classes `WebDAVAddFlowUITests`, `WebDAVConnectErrorUITests`, `WebDAVSeededUnlockUITests`, `CloudBrowserSmokeUITests`, `CloudUnlockSmokeUITests`, `CloudAccountEdgeUITests`: **6/6 passed**.
- Manual run in the simulator against the WebDAV mock: the switch filters in both directions, `vault.bin` opens and unlocks.
- `xcodebuild build-for-testing -scheme KeeForgeMac CODE_SIGNING_ALLOWED=NO`: builds.

Not run locally:

- `KeeForgeMacTests` and Mac UI tests (left to the `macos-unit-tests` CI job).
- The real Dropbox, OneDrive, WebDAV, and FTP servers; coverage is through the unit seams and the WebDAV UI-test double. `createFile`'s use of `includesAllFiles: true` and the Dropbox search cancellation check have no network-level test (there is no transport seam for either).

Notes for the UI test: on iOS 26 a plain synthesized tap did not reliably flip this switch (a real finger tap does), so the test presses for 0.15 s. It taps the row's file name because the existing plain-style row button only takes taps on its text. It also seeds the WebDAV account at launch instead of typing credentials, so iOS's "save password" prompt cannot steal focus mid-test.

# Checklist

- [x] Ran `xcodegen generate` after adding/removing/moving source files (no files added; project file unchanged)
- [x] Kept AutoFill extension target membership/imports in sync for shared code (no AutoFill files touched)
- [ ] Updated `KDBXCompatibilityTests` if parser/writer/save behavior changed (not applicable: no parser, writer, or save change)
- [x] Preserved accessibility identifiers or updated the affected UI tests (existing ids unchanged; new: `cloud.browser.show-all-files.toggle`, `cloud.browser.show-all-files.button`)
- [ ] Ran a `KeeForgeMacTests` slice when shared or Mac behavior changed (Mac target compiled; tests left to CI)


KeeForge/Services/Cloud/FTPCloudProvider.swift
```diff
@@ -567,13 +569,19 @@ final class FTPCloudProvider: CloudProvider, FTPConnecting, Sendable {
...
+    /// `.<name>.<UTC time-val>-<token>.<kind>`: hidden, and dropped from every
+    /// listing (`isScratchName`), even one showing all files. The time lets a
+    /// later save tell a leftover from a transfer in flight.
     static func scratchName(for name: String, kind: ScratchKind, createdAt: Date, token: String) -> String {
         ".\(name).\(timeVal(for: createdAt))-\(token).\(kind.rawValue)"
     }
 
+    /// Scratch files are this provider's own bookkeeping; one opened as a
+    /// database could be deleted by the stale-upload sweep.
+    static func isScratchName(_ fileName: String) -> Bool {
```
crazytan, line 581: Non-blocking: this predicate also hides legitimate dot-prefixed user files or folders that merely end in `.keeforge-upload` or `.keeforge-backup` (for example, `.vault.keeforge-backup`), even when Show All Files is enabled. Could this match the full generated timestamp/token scratch-name shape instead—as `creationDate(ofUploadScratchNamed:)` already does—and add a negative regression test?

crazytan (dismissed):
Approved. I left one non-blocking inline note about matching the complete FTP scratch-file name format.

